PR #13, opened 2019-02-12
Add cross-env package

eharris128:
### Issue
Current implementation does not support start and test-once scripts on windows. 

### Proposal
Add cross-env package and modify start and test-once scripts


fvictorio:
Thanks!

PR #17, opened 2019-02-15
Add wrap/unwrap weth component

fvictorio:
Part of #7, but styling is missing. I will merge this anyway because there are a lot of changes in other components (because of a refactor in the `Card` component) that might cause merge conflicts.


pablofullana:
Closes #7

PR #32, opened 2019-02-19
Add cancel order "x" button.

### LINKED ISSUE #22: Cancel Orders
As a logged in user, I should be able to cancel an order of mine.

unjapones:
Closes #22


src/store/actions.ts
```diff
@@ -157,3 +158,33 @@ export const initWallet = () => {
...
+export const cancelOrder = (order: SignedOrder) => {
+    return async (dispatch: any, getState: any) => {
+        const state = getState();
+        const ethAccount = getEthAccount(state);
+
+        const web3Wrapper = await getWeb3WrapperOrThrow();
+        const networkId = await web3Wrapper.getNetworkIdAsync();
+        const contractWrappers = new ContractWrappers(web3Wrapper.getProvider(), { networkId });
+        const tx = await contractWrappers.exchange.cancelOrderAsync(order, {
+            gasLimit: 1000000,
```
unjapones: @fvictorio I can move the repeated options object to `/src/constants`. Also, the options may be _optional_ (setup via an env variable) because I'm not 100% sure if the `out of gas` error has to do with using ganache. Have to test this on a test net.

fvictorio: Yes, let's do that.

src/store/actions.ts
```diff
@@ -157,3 +158,33 @@ export const initWallet = () => {
...
+        });
+        await web3Wrapper.awaitTransactionSuccessAsync(tx);
+
+        const relayer = getRelayer();
```
fvictorio: Maybe we can move the "load orders" logic to another action, and dispatch it both from here and from the init action?

src/components/marketplace/cancel_order_button.tsx
```diff
@@ -0,0 +1,68 @@
...
+    onCancelOrder: (order: SignedOrder) => Promise<any>;
+}
+
+type Props = OwnProps & DispatchProps;
+
+interface State {
+    isLoading: boolean;
+}
+
+const BUTTON = styled.button`
```
fvictorio: This should be `Button`.

src/components/marketplace/cancel_order_button.tsx
```diff
@@ -0,0 +1,68 @@
...
+    };
+
+    public render = () => {
+        const { isLoading } = this.state;
+        return <BUTTON type="button" disabled={isLoading} onClick={this._cancelOrder}>
+            <FontAwesomeIcon icon="times" />
+        </BUTTON>;
+    };
+
+    private readonly _cancelOrder  = async () => {
```
fvictorio: I'm not against this, but it would be nice to have a convention here. Should we use `private readonly` and an underscore prefix for all methods that are not lifecycle hooks? Pros: it makes sense. Cons: more keywords.

I'm inclined to say yes. And if we do it, maybe we could add a linter rule for that (if there is one).

src/components/marketplace/cancel_order_button.tsx
```diff
@@ -0,0 +1,68 @@
...
+    display: inline-block;
+    height: 1em;
+    outline: 0;
+
+    &:hover {
+        cursor: pointer;
+    }
+`;
+
+class CancelOrderButton extends React.Component<Props, State> {
```
fvictorio: Just a thought: It feels weird to have a container component used inside a presentational component. But I can't think of a downside of doing this (the parent component is still testable, if I'm not missing anything), and the alternative is passing down a callback for the cancel order, which is cumbersome.

unjapones: Feels a little weird but it encapsulates what the button does. We can refactor the base code to have presentational containers and presentational components. I'm not 100% sure, but having the latters splitted out may make UI customization easier.

src/components/marketplace/cancel_order_button.tsx
unjapones: In this case, I saw 2 linting errors:
* `Private member variable '_cancelOrder' is never reassigned; mark it as 'readonly'.`
* `private and protected members must be prefixed with an underscore`

fvictorio: Cool, so we only have to "remember" to make them private, since I found no rule that enforces making those methods private.

src/components/marketplace/cancel_order_button.tsx
fvictorio: Yeah, no need to change this, I'm OK with it.

src/store/actions.ts
unjapones: Done with this little refactor.

PR #46, opened 2019-02-20
(Feature) Adds default routing case

Agupane:
Adds a default route on the react router


src/index.tsx
```diff
@@ -25,6 +25,7 @@ const Web3WrappedApp = (
                     <Switch>
                         <Route exact={true} path="/" component={Marketplace} />
                         <Route exact={true} path="/my-wallet" component={MyWallet} />
+                        <Redirect to="/" />
```
fvictorio: This seems to work, but I don't think it's the idiomatic way to do it. I think you have to do something like:

```
  <Route component={Marketplace}/>
```

[`<Redirect>`](https://reacttraining.com/react-router/web/api/Redirect) redirects to the given path when that component is rendered. I believe this is working because `<Switch>` doesn't find any matching route and *then* renders the redirect.

PR #47, opened 2019-02-20
(Feature) Orderbook current list styling

### LINKED ISSUE #20: Revisit Orderbook (current) list styling

gabitoesmiapodo:
Closes #20 

It should look better now.

screen shot 2019-02-20 at 16 37 36
screen shot 2019-02-20 at 16 37 55
screen shot 2019-02-20 at 16 38 07

### DIFF
--- a/src/components/common/empty_content.tsx
+++ b/src/components/common/empty_content.tsx
@@ -1,12 +1,15 @@
 import React, { HTMLAttributes } from 'react';
 import styled from 'styled-components';
 
-interface Props extends HTMLAttributes<HTMLDivElement> {
+interface EmptyWrapperProps {
     alignAbsoluteCenter?: boolean;
+}
+
+interface Props extends HTMLAttributes<HTMLDivElement>, EmptyWrapperProps {
     text: string;
 }
 
-const EmptyContentWrapper = styled.div<Props>`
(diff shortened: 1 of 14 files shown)


src/components/common/empty_content.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { HTMLAttributes } from 'react';
+import styled from 'styled-components';
+
+interface Props extends HTMLAttributes<HTMLDivElement> {
+    alignAbsoluteCenter?: boolean;
```
fvictorio: The two usages of this component set this to `true`. Are we anticipating a component that needs to use this with `false`?

src/components/common/empty_content.tsx
```diff
@@ -0,0 +1,34 @@
...
+    text?: string;
```
fvictorio: Can we make this prop mandatory?

src/components/common/empty_content.tsx
gabitoesmiapodo: Yes, you might not need (or not want to) to center it occupying all of its container, or you may not have a parent with a `position: relative;` property (which could be disastrous).

src/components/common/empty_content.tsx
gabitoesmiapodo: `text?: string;` is mandatory now.

PR #48, opened 2019-02-20
(Feature) Fix total sizes

### LINKED ISSUE #42: Orderbook is not subtracting filled values
The total size for a price in the order book should not take into account the filled values.

For example, an order of size 100 with 30 filled, should add 70 to the total size for its price.

The same probably applies too to the "My size" column.

Agupane:
Now the orderbook sustracts the filled values
Closes #42


src/util/types.ts
```diff
@@ -63,12 +63,14 @@ export interface UIOrder {
     filled: BigNumber;
     price: BigNumber;
     status: OrderStatus;
+    emptySize: BigNumber;
```
fvictorio: This works but having a "denormalized" property might be error-prone and hurt maintainability, since this opens the door for having inconsistent states (where `emptySize + filled !== size`). Since this value can be computed from the others, I think it's best to just compute it when necessary. If this turns out to happen a lot, we can convert this type to a class with a getter method for `emptySize`.

And while we are on it, I'm not sure about the name `emptySize`. Maybe something like `available`?

src/util/types.ts
```diff
@@ -63,12 +63,14 @@ export interface UIOrder {
...
 }
 
 export interface OrderBookItem {
     side: UIOrderSide;
     size: BigNumber;
     price: BigNumber;
+    emptySize: BigNumber;
```
fvictorio: And here *maybe* we can just use size, since the size of an order book item is meant to be the sum of the available parts of the orders for that price. I'm not 100% sure about this, though.

src/util/ui_orders.ts
```diff
@@ -58,9 +58,17 @@ export const mergeByPrice = (orders: OrderBookItem[]): OrderBookItem[] => {
...
-        .map(order => ({
-            side: order.side,
-            price: order.price,
-            size: order.size,
-        }));
+        .map(order => {
+            /* Avoids Typescript Compilation on filled field */
+            const orderToChange: any = order;
+            const newSize =
+                orderToChange.size && orderToChange.filled
```
fvictorio: `orderToChange.filled` is not nullable, and by default it has 0, so this ternary doesn't seem necessary.

PR #49, opened 2019-02-20
Add Buy/Sell on limit

### LINKED ISSUE #23: Buy/sell (limit)
As a logged in user, I should be place a limit order (buy/sell).

- [ ] Buy limit
- [ ] Sell limit

fvictorio:
Closes #23. Depends on #45.

There are some pending things here that maybe can be done in a separate PR:

- The input is not validating that the value entered by the user is less than or equal to the amount of tokens they have.
- The "Order Details" section is missing, mainly because we don't have the USD price of ETH yet, but also because I don't know how this fee is supposed to be computed (is it the sum of the maker and taker fees?).

This PR also renames the `UIOrderSide` enum to `OrderSide`, since it's used everywhere, not just for UI stuff.


gabitoesmiapodo:
Design for reference:

screen shot 2019-02-21 at 15 48 15

src/util/orders.ts
```diff
@@ -0,0 +1,37 @@
...
+    tokenAddress: string;
+    wethAddress: string;
+    amount: BigNumber;
+    price: number;
+    exchangeAddress: string;
+}
+
+export const buildOrder = (params: BuildOrderParams, side: OrderSide): Order => {
+    const { account, exchangeAddress, amount, price, tokenAddress, wethAddress } = params;
+    const tomorrow = new BigNumber(Math.floor(new Date().valueOf() / 1000) + 3600 * 24);
```
unjapones: We can create a ticket later to adopt a constant value or something set in the environment (NNNN mins). Not really urgent right now.

fvictorio: Oh, yes, I totally forgot about this :man_facepalming: An environment variable seems the way to go.

unjapones: :+1: this should be a non-blocking detail. If you want, just create an issue to deal with it later.

Agupane:
The input fields allows to enter things like multiple "-" or ".", this behaviour is currently not generating problems with the contract interaction. It seems to avoid wrongs-inputs and change the values to undefined, the contract did not generated an error but we should keep it in mind (at least for user experience). According to what I have read, this could be a problem with the react input with type="number"

image
image
image

If you guys agree, we could open a new mini-issue to do some research later on.

@fvictorio 
@unjapones

fvictorio:
> The input fields allows to enter things like multiple "-" or "."

Yeah, that's just the HTML validation being permissive. The alternative is to make a text input with custom validation, but that's a deep rabbit hole. I guess we can do it at some other point if needed.

There are merge conflicts here, I will solve them now.

gabitoesmiapodo:
Styles for this can be reviewed.

I only styled the existing elements, and I will do the same when / if more exist.

You should see something like this now.

screen shot 2019-02-22 at 13 39 27
screen shot 2019-02-22 at 13 43 24
screen shot 2019-02-22 at 13 43 42

PR #54, opened 2019-02-21
(Feature) Lock tokens

### LINKED ISSUE #36: Allow locking the tokens
Right now tokens can only be unlocked. The user should be able to lock them back.

Agupane:
Let the users lock their tokens
Closes #36

### DIFF
--- a/src/store/actions.ts
+++ b/src/store/actions.ts
@@ -98,7 +98,6 @@ export const lockToken = (token: Token) => {
             if (tokenBalance.token.address !== token.address) {
                 return tokenBalance;
             }
-            /* We could check if the amount was correctly set before locking them (maybe) */
             return {
                 ...tokenBalance,
                 isUnlocked: false,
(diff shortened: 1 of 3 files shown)


src/store/actions.ts
```diff
@@ -77,6 +77,37 @@ export const unlockToken = (token: Token) => {
...
+        const tokenBalances = getTokenBalances(state);
+
+        const contractWrappers = await getContractWrappers();
+        /* This should be 0, cero amount of tokens should be available if we lock them */
+        const numberOfUnlockedTokens = new BigNumber('0');
+
+        await contractWrappers.erc20Token.setProxyAllowanceAsync(
+            token.address,
+            ethAccount,
+            numberOfUnlockedTokens,
```
fvictorio: I think it would be simpler to just inline the zero here. Even better, we could hide this behind some helper, but that's something we can do later.

Agupane: Done, now the zero is inline

PR #55, opened 2019-02-21
Add token icons

### LINKED ISSUE #11: Add token icons
As a user, in my account page, add token icons next to each token name.

Grab icons from: https://github.com/MetaMask/eth-contract-metadata

unjapones:
Connects #11.

* Refactor a little bit _known tokens_ information and initialization (adds file to `/src/common`, `/src/services` and `/src/util`).
* Add `TokenIcon` component to render an SVG, given a token symbol.

I might add some notes in the code for clarification.


unjapones:
If you manually test this and check it in the browser, no icon will be visible while on "My Wallet" because the SVGs' main color seems to be white.

This needs styling, or some way to colorize the SVG elements (0x Instant seems to use a `primary` field/attribute for each token's metadata):

background_purple

unjapones:
Another thing, we could have some kind of _fallback icon_ as seen in the designs (screenshot below): a component that renders the first 3 letters of the token's symbol with a coloured-circle background as in the mocks.

fallback

For now, `TokenIcon` renders the SVG, but it you think I should add this we can create a ticket or add it right into this PR.

fvictorio:
I like the idea of a fallback icon, and it shouldn't be too hard to do. @unjapones can you open an issue for that?

Agupane:
I agree with @fvictorio, fallback icons are a good Idea and should be considered in a new issue

src/services/known_tokens.ts
```diff
@@ -0,0 +1,44 @@
+import { WETH_TOKEN_SYMBOL } from '../common/constants';
+import { KNOWN_TOKENS_META_DATA } from '../common/tokens_meta_data';
+import {
+    getWethTokenFromTokensMetaDataByNetworkId,
+    mapTokensMetaDataToTokenByNetworkId,
+} from '../util/token_meta_data';
+import { Token } from '../util/types';
+
+export class KnownTokens {
```
fvictorio: Some thoughts on this.

First, can we move this to the `util` directory, based on what we talk about having pure/testable things there?

Second, it would be nice to make `KNOWN_TOKENS_META_DATA` a constructor parameters, so that this can be tested.

src/services/known_tokens.ts
```diff
@@ -0,0 +1,44 @@
...
+    private readonly _tokens: Token[] = [];
+    private readonly _wethToken: Token | null;
```
fvictorio: Maybe I'm missing something here, but I think there's no need to make this nullable. And the app shouldn't work without the weth token anyway.

src/services/known_tokens.ts
unjapones: Yup agree with both items. Will work on that now.

PR #66, opened 2019-02-25
(Feature) My size

### LINKED ISSUE #59: Add 'My Size' column to Orderbook

Agupane:
Add the column "My SIze" on the orderBook

Closes #59


src/components/marketplace/order_book.tsx
```diff
@@ -27,16 +28,30 @@ interface State {
...
+    selectedToken: Token,
+    mySizeOrders: OrderBookItem[] = [],
+) => {
     const size = tokenAmountInUnits(order.size, selectedToken.decimals);
     const price = order.price.toString();
     const priceColor = order.side === OrderSide.Buy ? themeColors.green : themeColors.orange;
     const time: string = '';
     const timeColor = time ? '#000' : themeColors.lightGray;
 
+    const mySize = mySizeOrders.reduce((sumSize, mySizeItem) => {
```
fvictorio: I don't think this will work if the user has more than one order with the same price, since it will only use the size of the last of those orders.

fvictorio: If this is indeed a bug, please add a proper failing test before fixing it (@mariano-aguero can probably help with that if needed).

fvictorio:
This looks good, just one minor thing I hadn't noticed before: when the "My Size" column is zero, it should show a dash (`-`) instead of `0.00`.

PR #68, opened 2019-02-26
(Feature) - Orderbook: Pad sizes with zeros 

### LINKED ISSUE #28: Orderbook details
- [x] Parent of #42 (Orderbook is not subtracting filled values)
- [x] Parent of #59 (Add 'My Size' column to Orderbook)
- [ ] Parent of #60 (Pad sizes with 0)

*More items to be added*

### LINKED ISSUE #60: Orderbook: Pad sizes with zeroes

mariano-aguero:
Closes #60 
Also closes #28 

- Added function to calculate the decimals to add to a given number
- Added tests
- Added component to show numbers with different colors

Should look like this now
screenshot_20190226_111844


src/util/number_utils.ts
```diff
@@ -0,0 +1,20 @@
...
+export const padRightSplitted = (numBg: BigNumber, decimals: number = 4): { num: string; diff: string } => {
+    const numBgToFixed = numBg.toFixed(decimals);
+    const numBgToString = numBg.toString();
+
+    const decimalPlaces = (numBgToString.split('.')[1] || []).length;
+
+    let diff = '';
+    let num = numBgToFixed;
+    if (!numBg.isZero() && decimalPlaces < decimals) {
+        diff = numBgToFixed.split(numBgToString).join('');
```
fvictorio: This is equivalent to `.replace(numBgToString, '')`, isn't it?

mariano-aguero: Yes @fvictorio , Do you think the replace is a better option?

mariano-aguero:
> Minor comment, but otherwise looks good. The only thing pending is for the numbers to be tabular. I think that can be done adding this CSS:
> 
> ```
> font-feature-settings: 'tnum' 1
> ```
> That's probably something for @gabitoesmiapodo, but if that's the only thing needed here, maybe we can just add it.

@fvictorio Do you want to create another issue? or we keep this issue open until @gabitoesmiapodo  solve it here

gabitoesmiapodo:
@mariano-aguero you have my blessing for this.

src/util/number_utils.ts
fvictorio: Yes

fvictorio:
> Do you want to create another issue? or we keep this issue open until @gabitoesmiapodo solve it here

Let's do it here.

src/util/number_utils.ts
mariano-aguero: Solved

gabitoesmiapodo:
@fvictorio @mariano-aguero 

This should be fixed now.

> Minor comment, but otherwise looks good. The only thing pending is for the numbers to be tabular. I think that can be done adding this CSS:

PR #70, opened 2019-02-26
(Feature) Order details in buy/sell component

### LINKED ISSUE #58: Add "Fee" field of "Order Details" in buy/sell component (when in Limit)
Use the maker and taker fees along with the order price to compute the total fee in ETH.

After this is done, we should add some mechanism for obtaining the price of ETH in USD.

- [ ] Show fee details in ETH
- [ ] Show fee details in USD

Agupane:
#Connects #58

**NOTE: disable AdBlock to manually test this feature, and increase the value of the maker fee constant to get a significant amount of USD.**

DIsplays fees amount (in 0x, USD) if the user creates an order (buy/sell limit)

It should look like so:

image
image


.env.example
```diff
@@ -0,0 +1,2 @@
+REACT_APP_MAKER_FEE=YOUR_MAKER_FEE
+REACT_APP_TAKER_FEE=YOUR_TAKER_FEE
```
fvictorio: This is a great idea. Two comments:

- Let's add the other variables: `REACT_APP_RELAYER_URL`, `REACT_APP_RELAYER_NETWORK_ID` and `REACT_APP_FEE_RECIPIENT` (I think I'm not forgetting anything).
- Let's use example values, so that it's easier to know what it's expected.

src/components/marketplace/buy_sell.tsx
```diff
@@ -87,8 +91,8 @@ const LabelContainer = styled.div`
     margin-bottom: 10px;
 `;
 
-const Label = styled.label`
-    color: #000;
+const Label = styled.label<{ color?: string }>`
+    color: ${props => (props.color ? props.color : '#000')};
```
fvictorio: Personal preference: `props.color || '#000'` is more idiomatic.

src/components/marketplace/buy_sell.tsx
```diff
@@ -203,26 +251,52 @@ class BuySell extends React.Component<Props, State> {
...
     public updatePrice: React.ReactEventHandler<HTMLInputElement> = e => {
         const price = parseFloat(e.currentTarget.value);
+        this.setState({ price }, () => this.updateOrderFee());
+    };
 
-        this.setState({ price });
+    public updateOrderFee = async () => {
+        const orderPrice = this.state.makerAmount.mul(this.state.price);
+        const orderFeeEther = orderDetailsFeeEther(orderPrice, this.state.tab);
+        const orderFeeDollar = await orderDetailsFeeDollar(orderPrice, this.state.tab);
```
fvictorio: Does this mean that we are going to make a request to the CMC API every key stroke? We may hit a rate limit if we do that.

Two ways of fixing this come to mind: we can periodically ask for the price and save it in the store, or we can add some caching to the function so that the price is not asked more than once every X minutes.

The first option is probably better, but the second one is easier. We can make the cache thing for now and add the price to the store in another PR.

fvictorio: More on this: manually calling `updateOrderFee` after `this.setState` doesn't look good. Since the fee is computed from the order values, I think it makes more sense to do it in the render function. But we can't do it there because this function is async. So more reason to just have the ETH price in the store. Again: no need to do it now, just something to keep in mind.

src/util/orders.test.ts
```diff
@@ -77,3 +76,54 @@ describe('buildOrder', () => {
         );
     });
 });
+describe('orderDetails', () => {
+    it('orderDetailsFeeEther should return a BigNumber ', () => {
```
fvictorio: Do we need this kind of tests if we have types?

src/util/orders.ts
```diff
@@ -35,3 +38,33 @@ export const buildOrder = (params: BuildOrderParams, side: OrderSide): Order =>
...
+        fee = new BigNumber(MAKER_FEE);
+    }
+
+    if (orderType === OrderSide.Sell) {
+        fee = new BigNumber(TAKER_FEE);
+    }
+
+    // Fee is a percentage
+    const orderPriceConverted = tokenAmountInUnits(orderPrice, tokenDecimals);
+    return fee.mul(orderPriceConverted).div(100);
```
fvictorio: This doesn't seem correct. Here's how I understand the fees computation.

Both in buy and sell orders, you have a maker and a taker amount. You multiply them by the maker fee and the taker fee, respectively. This will give you an amount in ETH and the other in the selected token. Which is which depends on the type of the order: in sell orders, the resulting taker fee will be in ETH; in buy orders, the maker fee will be in ETH. So you multiply the other value (the one in units of the selected token) by the price, to convert it to ETH. Finally you sum both values: that's the total fee.

Does this make sense?

src/util/orders.test.ts
Agupane: I deleted this and added a better test that checks a result against a mocked dolar vallue

src/components/marketplace/buy_sell.tsx
Agupane: I implemented a cache with interval, we could maybe create another issue to move all this things to the store directly

src/components/marketplace/buy_sell.tsx
Agupane: Done

.env.example
Agupane: I added them, also I'd added an env variable to set the cache interval

src/components/marketplace/buy_sell.tsx
fvictorio: I opened #77.

src/components/marketplace/buy_sell.tsx
```diff
@@ -203,26 +251,59 @@ class BuySell extends React.Component<Props, State> {
...
     public updatePrice: React.ReactEventHandler<HTMLInputElement> = e => {
         const price = parseFloat(e.currentTarget.value);
+        this.setState({ price }, () => this.updateOrderFee());
+    };
 
-        this.setState({ price });
+    public updateOrderFee = async () => {
+        const orderFeeEther = orderDetailsFeeEther(
+            this.state.makerAmount,
+            new BigNumber(this.state.price),
```
fvictorio: Am I missing something or this function expects to receive the taker amount, not the price?

src/util/orders.test.ts
```diff
@@ -1,8 +1,8 @@
 import { BigNumber } from '0x.js';
 
-import { OrderSide } from '../util/types';
-
-import { buildOrder } from './orders';
+import * as dolarUtils from './market_prices';
```
fvictorio: Typo, `dollar`

src/util/orders.test.ts
```diff
@@ -77,3 +77,37 @@ describe('buildOrder', () => {
...
 });
+describe('orderDetails', () => {
+    it('orderDetailsFeeEther should calculate totalFee ', () => {
+        // given
+        const makerAmount = new BigNumber(5000000000000000000);
+        const takerAmount = new BigNumber(500000000000000000);
+        const orderType = OrderSide.Buy;
+
+        // when
+        const orderInEther = orderDetailsFeeEther(makerAmount, takerAmount, orderType);
```
fvictorio: It seems like this function (and therefore these tests) has a implicit dependency on the environment variables.

This is what I would do: make the function accept the maker/taker fee values, but use the environment variables as default arguments. That way you don't have to change the code that uses the function but you can be more explicit in the tests.

src/util/orders.ts
```diff
@@ -35,3 +38,52 @@ export const buildOrder = (params: BuildOrderParams, side: OrderSide): Order =>
...
+    tokenDecimals: number = 18,
+): BigNumber => {
+    let totalFee = new BigNumber(1);
+
+    // Convert the makerAmount
+    const makerAmountConverted = new BigNumber(tokenAmountInUnits(makerAmount, tokenDecimals));
+    if (orderType === OrderSide.Buy) {
+        // Calculate makerFee
+        const makerFeeWithoutPrice = makerAmountConverted.mul(MAKER_FEE);
+        const makerFeeWithPrice = makerFeeWithoutPrice.mul(takerAmount.div(makerAmountConverted));
```
fvictorio: Is this correct? When you create a buy order, the maker asset is wETH, so it shouldn't be multiplied by the price.

src/util/orders.test.ts
fvictorio: Also, please add at least one test for a buy order and another for a sell order.

.env.example
```diff
@@ -1,2 +1,6 @@
-REACT_APP_MAKER_FEE=YOUR_MAKER_FEE
-REACT_APP_TAKER_FEE=YOUR_TAKER_FEE
+REACT_APP_MAKER_FEE=YOUR_MAKER_FEE (ex 1.0)
+REACT_APP_TAKER_FEE=YOUR_TAKER_FEE (ex 0.1)
+REACT_APP_RELAYER_URL=YOUR_RELAYER_URL (ex  'http://localhost:3001/api/v2')
+REACT_APP_RELAYER_NETWORK_ID=RELAYER_NETWORK_ID (ex 50)
+REACT_APP_FEE_RECIPIENT=ADDRESS_OF_THE_FEE_RECIPIENT (ex '0x0000000000000000000000000000000000000000')
+REACT_APP_CACHE_CHECK_INTERVAL=TIME_TO_RELOAD_CACHE (Ex 60000 for one minute)
```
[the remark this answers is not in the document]
Agupane: Done

src/util/orders.ts
mariano-aguero: The conditional logic was modified. Can you check it ? Thanks

src/components/marketplace/buy_sell.tsx
mariano-aguero: The price is the wETH that the user expects to obtain from the sale. Both are ERC20 (ZRX is the makerAmount and the wETH is the takerAmoun). Please correct me if it's wrong.
Regards

fvictorio: No, that's not how it works. The maker and taker amount are the things that are being traded. The price is the relation between them. The only case where the price is equal to the taker amount (in the case of a sell) is when you are selling exactly 1 token.

src/util/orders.test.ts
```diff
@@ -77,3 +77,55 @@ describe('buildOrder', () => {
         );
     });
 });
+describe('orderDetails', () => {
+    it('orderDetailsFeeEther should calculate totalFee for buy order ', () => {
```
fvictorio: The test descriptions are inverted.

src/util/orders.test.ts
```diff
@@ -77,3 +77,55 @@ describe('buildOrder', () => {
...
+        const takerAmount = new BigNumber(500000000000000000);
+        const orderType = OrderSide.Sell;
+        const MAKER_FEE = '0.1';
+        const TAKER_FEE = '0.05';
+
+        // when
+        const orderInEther = orderDetailsFeeEther(makerAmount, takerAmount, orderType, MAKER_FEE, TAKER_FEE);
+
+        // then
+        const resultExpected = new BigNumber(75000000000000000);
```
fvictorio: According to my (manual) calculations, this should also be 0.75. In fact, if you look at the numbers you are using, it makes no sense for this fee to be 10^18 times bigger than the next scenario.

src/util/orders.ts
fvictorio: Now it _looks_ correct, but one of the tests seems wrong.

unjapones: Hey guys, sorry to jump-in late in the thread.

I believe we just have to convert the `makerFee` value (wich is always in ZRX) to ETH, without involving the order's asset amounts, price, and/or the `takerFee`.

So... the way I see things, the *Total Cost* showed to the user should only involve the `makerFee`: i.e. as a user, I'm mostly interested in "how much _I have to pay_". The `takerFee` will be the taker's thing to deal with.

I'm not 100% sure of this, we can confirm this with the 0x team tomorrow.

src/util/orders.test.ts
Agupane: now it's updated

src/util/orders.test.ts
Agupane: Updated

src/components/marketplace/buy_sell.tsx
```diff
@@ -151,9 +151,11 @@ const TokenTextButtonUppercase = styled.span`
 class BuySell extends React.Component<Props, State> {
     public state = {
         makerAmount: new BigNumber(0),
+        orderFeeEther: new BigNumber(0),
```
fvictorio: Is this still used?

unjapones: @Agupane yup, it looks like the fee-handling code is not needed anymore here: attributes in the component's state and `updateOrderFee` method (since everything is encapsulated in `OrderDetails`).

src/components/marketplace/buy_sell.tsx
```diff
@@ -217,25 +219,38 @@ class BuySell extends React.Component<Props, State> {
                             </FieldContainer>
                         </>
                     )}
-                    <Button theme="secondary" onClick={tab === Tab.Buy ? this.buy : this.sell}>
-                        {tab === Tab.Buy ? 'Buy' : 'Sell'}{' '}
+                    {orderType === OrderType.Limit ? <OrderDetails /> : null}
```
unjapones: Minor coment: probably the `<OrderDetails />` can be included at the end fragment above, below the closing `</FieldContainer>` in line 219.

src/util/orders.ts
```diff
@@ -83,3 +86,62 @@ export const buildMarketOrders = (
 
     return [ordersToFill, roundedAmounts, canBeFilled];
 };
+
+export const orderDetailsFeeEther = (
```
unjapones: Now that we moved the responsability of displaying the fees in ZRX/USD to a component (since it is something the UI does), do we need this method? I think we no longer use it and we can remove it + its corresponding tests.

src/components/marketplace/buy_sell.tsx
Agupane: now it's removed

src/util/orders.ts
Agupane: now it's removed

PR #73, opened 2019-02-26
(Feature) - Capitalize tokens' short name

### LINKED ISSUE #72: Capitalize tokens' short name
As a user, I should see every token (but wETH) short name capitalized, everywhere.

Example:

mariano-aguero:
Closes #72 

* Apply capitalize to symbol on getSelectedToken function

Should look like this now

screenshot_20190226_175310

### DIFF
--- a/src/components/marketplace/buy_sell.tsx
+++ b/src/components/marketplace/buy_sell.tsx
@@ -135,12 +135,25 @@ const TokenContainer = styled.div`
     z-index: 12;
 `;
 
-const TokenText = styled.span`
+const tokenTextStyle = `
     color: #333;
     font-size: 14px;
     font-weight: normal;
     line-height: 21px;
     text-align: right;
+`;
+
+const TokenText = styled.span`
+    ${tokenTextStyle}
+`;
+
+const TokenTextUppercase = styled.span`
+    ${tokenTextStyle}
+    text-transform: uppercase;
+`;
+
+const TokenTextButtonUppercase = styled.span`
+    text-transform: uppercase;
 `;
 
 class BuySell extends React.Component<Props, State> {
@@ -191,7 +204,7 @@ class BuySell extends React.Component<Props, State> {
                             value={makerAmount}
                         />
                         <TokenContainer>
-                            <TokenText>{selectedTokenSymbol}</TokenText>
+                            <TokenTextUppercase>{selectedTokenSymbol}</TokenTextUppercase>
                         </TokenContainer>
                     </FieldContainer>
                     <LabelContainer>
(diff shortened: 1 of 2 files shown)


unjapones:
@pablofullana I believe the name of tokens should have all of its symbol's letters in uppercase, instead of _only_ the first letter. The exception to this rule seems to be `wETH` :confused:.

src/store/selectors.ts
```diff
@@ -11,7 +11,14 @@ export const getEthBalance = (state: StoreState) => state.blockchain.ethBalance;
 export const getWethBalance = (state: StoreState) => state.blockchain.wethBalance;
 export const getOrders = (state: StoreState) => state.relayer.orders;
 export const getUserOrders = (state: StoreState) => state.relayer.userOrders;
-export const getSelectedToken = (state: StoreState) => state.relayer.selectedToken;
+export const getSelectedToken = (state: StoreState) => {
+    const selectedToken = state.relayer.selectedToken;
```
unjapones: @mariano-aguero I would prefer not to transform the returned value of this function: what happens if we want to check for equality between `getSelectedToken(state).symbol` with a given `token.symbol`? (we will probably have to convert to uppercase the 2nd one every time we do this).

Should we do this on the React component instead? We can use JS `string.toUpperCase()` or CSS `text-transform: uppercase` accordingly.

We can also change the tokens' metadata file and set the symbols in upper case (and `wETH` with its special-casing :confused:). For this 2nd option, probably we should check with @fvictorio and @Agupane since I may be missing something

fvictorio: Yes, this should be done in the component, it's totally a view concern.

We could add it to the metadata... but just uppercasing the symbol should be enough for now. We are already handling `wETH` as a special case everywhere, so doing it once more here for the label shouldn't be a big deal.

Agupane: According to what I saw, this could be a good choice,  I made some tests and seems OK. but I'm not 100% sure if this could generate some unexpected behaviour.

fvictorio: Plus, this seems to be mutating the store, doesn't it? Shouldn't `readonly` prevent us from doing that, or the `readonly`ness is lost when you copy a value?

fvictorio: Nevermind, `readonly` is just applied at the top level. The point remains: we shouldn't mutate the store directly.

mariano-aguero: The store was modified to its previous state, , and a CSS transformation was added on the React component.

src/components/marketplace/buy_sell.tsx
```diff
@@ -135,14 +135,27 @@ const TokenContainer = styled.div`
...
 `;
 
+const TokenText = styled.span`
+    ${tokenTextStyle}
+`;
+
+const TokenTextUppercase = styled.span`
+    ${tokenTextStyle}
+    text-transform: uppercase;
+`;
```
fvictorio: This can be written as:

```javascript
const TokenText = styled.span`
    color: #333;
    font-size: 14px;
    font-weight: normal;
    line-height: 21px;
    text-align: right;
`;

const TokenTextUppercase = styled(TokenText)`
    text-transform: uppercase;
`;
```

I haven't used `styled-components` enough to assure that this is more idiomatic, but it's shorter and (IMO) clearer.

I'm going to approve it anyway because it's not reason enough to block the PR.

mariano-aguero: Done!

PR #76, opened 2019-02-27
(Feature) Refactored Price Field

### LINKED ISSUE #67: Use a BigNumberInput for the price field
We are using a `type="number"` input, which can cause precision errors. Let's change it for a `BigNumberInput`. Also, if there are places where the price is represented as a number, they should be changed to `BigNumber`.

Agupane:
closes #67 

Refactors the price field input with a BigNumber variable


fvictorio:
This seems to share a commit with #74, although I see no reason for branching from that PR.

Not a big deal because it's only one commit, just something to keep in mind.

PR #81, opened 2019-02-28
(Feature) Automatic ui updates

### LINKED ISSUE #75: Automatic update UI on account/network changed
When the user go to metamask and change the current account or the current network, the UI should be re-rendered based on that information

Agupane:
closes #75 

Performs an ui update once the user changes the current account or the current network.

If the user changes the account the ui is re-rendered with the necesary values but not reloaded. If the user changes the network the page is reloaded in order to avoid problems with RC20TokenContract api:

image
image


Agupane:
Closes this to make modifications based on feedback, new pr on: #84

PR #83, opened 2019-02-28
(Feature / static) Wallet dropdown

### LINKED ISSUE #63: (style / static) Wallet dropdown
Children of #61 

screen shot 2019-02-22 at 16 04 21

gabitoesmiapodo:
Closes #63 

This is an easy one, I think.

- "Copy address to clipboard" should work.
- The remaining elements just display a message for now.

It should look like this.

screen shot 2019-02-28 at 14 21 31

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
         "history": "^4.7.2",
         "http-proxy-middleware": "^0.19.1",
         "react": "^16.7.0",
+        "react-copy-text": "^1.0.1",
         "react-dom": "^16.7.0",
         "react-modal": "^3.8.1",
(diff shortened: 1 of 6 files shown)


fvictorio:
The "Copy to clipboard" only works the first time, since the second time it sets the state to the same value and so the render is not triggered.

There seems to be some better alternatives to do this:

- https://www.npmjs.com/package/react-clipboard.js
- https://www.npmjs.com/package/react-copy-to-clipboard

The first one looks better to me (less downloads but it seems more maintained).

gabitoesmiapodo:
@fvictorio 

Should work now.

PR #85, opened 2019-02-28
(Feature) - Add fallback token icon/component

### LINKED ISSUE #56: Add fallback token icon/component
Check the comments around: https://github.com/0xProject/0x-launch-kit-frontend/pull/55#issuecomment-466194725

mariano-aguero:
Closes #56 

it should look like this, if the icon does not exist

screenshot_20190228_160006


src/components/common/icons/token_icon.tsx
```diff
@@ -69,9 +70,39 @@ export const TokenIcon = (props: Props) => {
     const { token } = props;
     const TokenIconComponentName = getTokenIconNameBySymbol(token.symbol) as keyof typeof TokenIcons;
     const Icon: React.FunctionComponent = TokenIcons[TokenIconComponentName];
-    return Icon ? React.createElement(Icon, null, null) : null;
+    return Icon
+        ? React.createElement(Icon, null, null)
+        : fallbackIcon({
+              fill: (token && token.primaryColor) || randomColor(),
```
unjapones: @mariano-aguero : can we get rid of the `randomColor()` ? I may become a little odd to the user if the color of the icon for a token is different every time the browser is refreshed.

Also `primaryColor` is now part of the corresponding TS interface: I imagine this will encourage the addition of Tokens' Metadata with that info.

Worst case scenario, use a default constant and boring-grey color.

src/common/constants.ts
```diff
@@ -15,3 +15,5 @@ export const FEE_RECIPIENT = process.env.REACT_APP_FEE_RECIPIENT || ZERO_ADDRESS
 
 export const MAKER_FEE = process.env.REACT_APP_MAKER_FEE || '0';
 export const TAKER_FEE = process.env.REACT_APP_TAKER_FEE || '0';
+
+export const DEFAULT_FALLBACK_ICON_COLOR = '#808080';
```
fvictorio: @gabitoesmiapodo Would it make sense to make this part of the `theme` variables?

src/components/common/icons/token_icon.tsx
```diff
@@ -69,9 +70,31 @@ export const TokenIcon = (props: Props) => {
...
+              fill: (token && token.primaryColor) || DEFAULT_FALLBACK_ICON_COLOR,
+              title: token.symbol && token.symbol.toUpperCase(),
+          });
 };
 
 const getTokenIconNameBySymbol = (symbol: string): string => {
     return `${symbol.charAt(0).toUpperCase()}${symbol.slice(1)}TokenIcon`;
 };
+
+const fallbackIcon = ({
```
fvictorio: Can we convert this into a `FallbackIcon` component? I think we just have to change its name and how it's used above.

src/common/constants.ts
gabitoesmiapodo: If it's going to be used elsewhere, yes it would make sense.

src/components/common/icons/token_icon.tsx
gabitoesmiapodo: > I may become a little odd to the user if the color of the icon for a token is different every time the browser is refreshed.

Is that so? I mean, a few apps and sites do that when they don't have a stablished color / icon / image for some elements.

This Whatsapp's user names, for example. Or Google's apps and its avatars.

mariano-aguero: Done

src/components/common/icons/token_icon.tsx
mariano-aguero: Done

mariano-aguero:
screenshot_20190301_173954
screenshot_20190301_173940

mariano-aguero:
@gabitoesmiapodo  Can you center the icon please ? Thanks

src/common/constants.ts
mariano-aguero: Done!

gabitoesmiapodo:
> @gabitoesmiapodo Can you center the icon please ? Thanks

No, it should be done here: https://github.com/0xProject/0x-launch-kit-frontend/issues/19

PR #86, opened 2019-03-01
Use maker account as senderAddress

fvictorio:
The [spec](https://github.com/0xProject/0x-protocol-specification/blob/master/v2/v2-specification.md#contract-interactions) says that the sender address is:

> Address that is allowed to call Exchange contract methods that affect this order. If set to 0, any address is allowed to call these methods.

We are setting it to 0 right now, which of course is not a good idea (it would mean, among other things, that anyone can cancel the order).

This PR uses the maker of the order as the sender address.


fvictorio:
Wait, nevermind, this will also mean that only the maker can fill the order, and my comment about the cancel seems to be wrong (or at least it should be).

PR #89, opened 2019-03-01
(Feature) Static notifications dropdown

### LINKED ISSUE #61: (style / static) Add header's missing elements
Parent of #62 
Parent of #63 
Parent of #64  
Parent of #65

launch_kit_ui_erc20_ _figma

### LINKED ISSUE #64: (style / static) Notifications dropdown
Children of #61 

screen shot 2019-02-22 at 16 04 12

gabitoesmiapodo:
Closes #64 
Closes #61 

- Dropdown has no tip on top (the little triangle thing), because it's imposible to align it with the bell icon and keep the dropdown completely inside the screen.
- Little red circle for unread / new notifications appears on first load, it should disappear when you open the dropdown. Reload to make it appear again. 
- No progress bar for active items (for now, at least).
- Items are not clickable (this can be changed easily later).

screen shot 2019-03-01 at 12 03 55

### DIFF
--- a/src/components/account/notifications_dropdown.tsx
+++ b/src/components/account/notifications_dropdown.tsx
@@ -1,5 +1,5 @@
 import React, { HTMLAttributes } from 'react';
-import styled from 'styled-components';
+import styled, { keyframes } from 'styled-components';
 
 import { themeColors, themeDimensions } from '../../util/theme';
 import { CardBase } from '../common/card_base';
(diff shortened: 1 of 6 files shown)


fvictorio:
Can we make it so that the notifications window only closes when you click outside of it? Right now you can't select the text inside it, for example, because it closes.

This is extremely low priority, I'm only asking in case it's a very low effort change.

gabitoesmiapodo:
@fvictorio 

All right, done.

PR #95, opened 2019-03-06
(Feature) Realtime updates

### LINKED ISSUE #30: Realtime updates v1
As a user, I'd like to see prices, sizes, statuses updated in realtime, so I don't have to manually refresh the page.

This first iteration of this feature aims to implement this real time updates via polling.

Open for discussion

Agupane:
Implements realtime updates of the UI using polling
closes #30


src/store/actions.ts
```diff
@@ -290,3 +290,24 @@ export const submitMarketOrder = (amount: BigNumber, side: OrderSide) => {
...
+
+            const knownTokens = getKnownTokens(networkId);
+
+            const tokenBalances = await Promise.all(
+                knownTokens.getTokens().map(token => tokenToTokenBalance(token, ethAccount)),
+            );
+
+            dispatch(getAllOrders());
+            dispatch(getUserOrders());
+            dispatch(setTokenBalances(tokenBalances));
```
fvictorio: Can we also update the eth and weth balance?

src/components/app.tsx
```diff
@@ -3,21 +3,36 @@ import { connect } from 'react-redux';
...
 
 class App extends React.Component<Props> {
+    public state = {
+        isActiveCheckUpdates: false,
+    };
+
     public componentWillMount = () => {
         this.props.onInitWallet();
+        /* Enables realtime updates of the store using pooling */
+        if (!this.state.isActiveCheckUpdates) {
```
fvictorio: This looks a little odd to me. First, because using the state of the component for this seems like a stretch. And second because `componentWillMount` should only be called once anyway.

I think a better approach is something like this:

```typescript
if (!this.updateStoreInterval) {
  this.updateStoreInterval = setInterval(...);
}
```

and you can also clear the interval in `componentWillUnmount` (probably not necessary since this is the root, but a good practice anyway). You'll have to declare `updateStoreInterval` as a field of the class; I'm not sure about its type but it should be easy to find out.

src/store/actions.ts
Agupane: it's done

src/components/app.tsx
Agupane: it's done

PR #97, opened 2019-03-06
Add StepsModal and basic sign_order_step.tsx.

### LINKED ISSUE #93: Add base modal flow mechanism
Add base modal to implement user interaction flows (e.g: "Create order with no WETH and no unlocked ZRX", "Create order with no ZRX", etc).

Assume the first flow is going to be simple: "Sign Order + submit to the relayer backend" (assume WETH and ZRX are unlocked + enough balance).

unjapones:
Closes #93.

### Includes:

* Add basic StepsModal UI component to render _ flow steps_ (check definition in `types.ts`) + store/actions/selectos updates to handle them accordingly.
** First _flow_ creation is handled in `actions`, but since there is only one significant step, there is no much to describe.
* Basic SignOrder step with _internal statuses_ like `Initial`, `Loading`, `Done`: this may be generalized (or moved to the store) when we implement more steps.

### Does not include:

* Styling: the idea is to provide a mechanisim so styling tasks can be performed without having to go through each step and its corresponding _status_.


src/store/reducers.ts
```diff
@@ -56,9 +63,47 @@ export function relayer(state: RelayerState = initialRelayerState, action: RootA
...
+            return { ...state, isVisible: action.payload };
+        case getType(actions.setStepsModalDoneSteps):
+            return { ...state, doneSteps: action.payload };
+        case getType(actions.setStepsModalPendingSteps):
+            return { ...state, pendingSteps: action.payload };
+        case getType(actions.setStepsModalCurrentStep):
+            return { ...state, currentStep: action.payload };
+        case getType(actions.stepsModalAdvanceStep):
+            const { doneSteps, currentStep, pendingSteps } = state;
+            if (pendingSteps.length === 0 && currentStep !== null) {
```
unjapones: `@TODO`: take this out as different function and add tests.

unjapones:
BTW: the PR does not contain code that will allow a super easy styling (like, for example, "_setting something on the initial store's state so a particual view is visible on the flow modal_"). 

After reviewing and/or merging this, we can probably discuss it.

src/util/types.ts
```diff
@@ -42,10 +42,31 @@ export interface RelayerState {
...
+    kind: StepKind.BuySellLimit;
+    amount: BigNumber;
+    price: BigNumber;
+    side: OrderSide;
+}
+
+export type Step = StepBuySellLimitOrder;
+
+export interface StepsModalState {
+    readonly isVisible: boolean;
```
fvictorio: What do you think about doing this a computed property through a selector? Maybe the modal is visible when the current step is not null?

src/util/types.ts
```diff
@@ -42,10 +42,31 @@ export interface RelayerState {
...
+    readonly doneSteps: Step[];
+    readonly currentStep: Step | null;
+    readonly pendingSteps: Step[];
+}
+
 export interface StoreState {
     readonly router: RouterState;
     readonly blockchain: BlockchainState;
     readonly relayer: RelayerState;
+    readonly stepsModal: StepsModalState;
```
fvictorio: I'm working on the notifications feature, and I added a `UIState` that holds the part of the global state that is related to the UI. Do you mind doing it that way so we don't have conflicts here? I guess stepsModal should be a property in `UIState`.

src/store/reducers.ts
```diff
@@ -56,9 +63,47 @@ export function relayer(state: RelayerState = initialRelayerState, action: RootA
...
+                return {
+                    ...state,
+                    doneSteps: doneSteps.concat([currentStep as Step]),
+                    currentStep: null,
+                };
+            } else if (pendingSteps.length > 1) {
```
fvictorio: I *think* you can make this just the `else` and remove the block below. It seems to me they do the same thing when `pendingSteps.length === 1`. Am I missing something?

src/store/reducers.ts
```diff
@@ -56,9 +63,47 @@ export function relayer(state: RelayerState = initialRelayerState, action: RootA
...
+                return {
+                    ...state,
+                    pendingSteps: pendingSteps.slice(1, pendingSteps.length),
```
fvictorio: This is just personal opinion, but I prefer `a.slice(1)` over `a.slice(1, a.length)`.

src/store/actions.ts
```diff
@@ -229,17 +257,39 @@ export const cancelOrder = (order: SignedOrder) => {
     };
 };
 
-export const submitLimitOrder = (amount: BigNumber, price: BigNumber, side: OrderSide) => {
-    return async (dispatch: any, getState: any) => {
+export const resetSteps = () => {
```
fvictorio: Pretty sure that this could be a "normal" (non-thunk) action:

```typescript
export const resetSteps = createAction('RESET_STEPS');
```

that then you just handle in the reducer by setting pending/done steps to `[]` and current step to `null`.

src/store/reducers.ts
unjapones: Oh yeah, I keep forgetting about this optional argument :|

src/store/actions.ts
unjapones: :+1:  will do that and return the initial state in the reducer.

src/store/actions.ts
```diff
@@ -61,6 +71,26 @@ export const setSelectedToken = createAction('SET_SELECTED_TOKEN', resolve => {
     return (selectedToken: Token | null) => resolve(selectedToken);
 });
 
+export const setStepsModalVisibility = createAction('SET_STEPSMODAL_VISIBILITY', resolve => {
```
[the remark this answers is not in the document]
unjapones: Oh yeah I forgot that one.

PR #100, opened 2019-03-07
(Feature) AdBlock fallback

### LINKED ISSUE #82: Add addblock fallback
If a user has addblock / addblock plus installed on the browser, when the app tries to get the eth USD value, the following error is displayed:
image

This should be solved or at least detect the plugin and display a message

Agupane:
Adds an alert if the user has addblock to tell him that he should disable it to use the dApp
it should look like so:
image

This will need probably some styling

closes #82 

## Updated 11/03/2019
As discussed on the comments, this will need styles for displaying the adblock fallback message, the modal should look like the issue #118, a new [issue](https://github.com/0xProject/0x-launch-kit-frontend/issues/127) has been open for this

### DIFF
--- a/src/components/common/adblock_detector.tsx
+++ b/src/components/common/adblock_detector.tsx
@@ -4,29 +4,29 @@ interface Props extends HTMLAttributes<HTMLButtonElement> {}
 
 interface State {}
 
-class AddBlockDetector extends React.Component<Props, State> {
+class AdBlockDetector extends React.Component<Props, State> {
     constructor(props: Props) {
         super(props);
     }
 
     public componentDidMount = async () => {
-        const isAddBlockDetected = await this.detectAddBlock();
+        const isAddBlockDetected = await this.detectAdBlock();
         if (isAddBlockDetected) {
             alert('Ad-Block DETECTED, please disable it in order to use our dApp!');
             window.location.reload();
         }
     };
 
-    public detectAddBlock = () => {
+    public detectAdBlock = () => {
         return new Promise<boolean>((resolve, reject) => {
             /* Creates a bait for ad block **/
             const elem = document.createElement('div');
             elem.className = 'adclass';
             document.body.appendChild(elem);
-            let isAddBlockDetected;
+            let isAdBlockDetected;
             window.setTimeout(() => {
-                isAddBlockDetected = !(elem.offsetWidth || elem.offsetHeight || elem.getClientRects().length);
-                resolve(isAddBlockDetected);
+                isAdBlockDetected = !(elem.offsetWidth || elem.offsetHeight || elem.getClientRects().length);
+                resolve(isAdBlockDetected);
             }, 0);
         });
     };
@@ -36,4 +36,4 @@ class AddBlockDetector extends React.Component<Props, State> {
         return children;
     };
 }
-export { AddBlockDetector };
+export { AdBlockDetector };
--- a/src/components/common/adblock_detector.tsx
+++ b/src/components/common/adblock_detector.tsx
@@ -1,19 +1,13 @@
-import React, { HTMLAttributes } from 'react';
+import React from 'react';
 
-interface Props extends HTMLAttributes<HTMLButtonElement> {}
-
-interface State {}
-
-class AdBlockDetector extends React.Component<Props, State> {
-    constructor(props: Props) {
-        super(props);
-    }
-
+/* TODO -- This could be refactored using Hooks **/
+class AdBlockDetector extends React.Component {
     public componentDidMount = async () => {
         const isAddBlockDetected = await this.detectAdBlock();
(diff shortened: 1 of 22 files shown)


src/pages/marketplace.tsx
```diff
@@ -10,15 +11,17 @@ class Marketplace extends React.PureComponent {
...
-                <Sidebar>
-                    <BuySellContainer />
-                </Sidebar>
-                <MainContent>
-                    <OrderHistoryContainer />
-                </MainContent>
-                <Sidebar>
-                    <OrderBookTableContainer />
-                </Sidebar>
+                <AdBlockDetector>
```
fvictorio: Hmm, I don't see why this should wrap the other components; just having it "beside" them should be enough. @unjapones what do you think?

src/components/common/adblock_detector.tsx
```diff
@@ -0,0 +1,39 @@
...
+class AdBlockDetector extends React.Component<Props, State> {
+    constructor(props: Props) {
+        super(props);
+    }
+
+    public componentDidMount = async () => {
+        const isAddBlockDetected = await this.detectAdBlock();
+        if (isAddBlockDetected) {
+            alert('Ad-Block DETECTED, please disable it in order to use our dApp!');
+            window.location.reload();
```
fvictorio: I think we should remove the reload and show a message like "We detected you are using an ad blocker. Keep in mind that this dapp may not work correctly with it enabled". @pablofullana thoughts?

yarn.lock
```diff
@@ -1889,6 +1889,11 @@ acorn@^6.0.1, acorn@^6.0.2, acorn@^6.0.4:
   resolved "https://registry.yarnpkg.com/acorn/-/acorn-6.0.5.tgz#81730c0815f3f3b34d8efa95cb7430965f4d887a"
   integrity sha512-i33Zgp3XWtmZBMNvCr4azvOFeWVw1Rk6p3hfi3LUDvIFraOMywb1kAtrbi+med14m4Xfpqm3zRZMT+c0FNE7kg==
 
+adblock-detect@^1.0.5:
```
unjapones: Is this alright? I don't see any `package.json` update.

src/pages/marketplace.tsx
unjapones: Yup, no need to wrap any content here. We can probably add it to [`/src/index.tsx`](https://github.com/0xProject/0x-launch-kit-frontend/blob/development/src/index.tsx#L25) as a sibling of `GeneralLayout`, i.e.:
```jsx
            <AppContainer>
                <AdBlockDetector/>
                <GeneralLayout>
                    <Switch>

```

src/components/common/adblock_detector.tsx
```diff
@@ -0,0 +1,39 @@
...
+            document.body.appendChild(elem);
+            let isAdBlockDetected;
+            window.setTimeout(() => {
+                isAdBlockDetected = !(elem.offsetWidth || elem.offsetHeight || elem.getClientRects().length);
+                resolve(isAdBlockDetected);
+            }, 0);
+        });
+    };
+
+    public render = () => {
```
unjapones: Since we probably don't want to use the component to wrap anything, this can be changed to:
```typescript
public render = () => null;
```

src/components/common/adblock_detector.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { HTMLAttributes } from 'react';
+
+interface Props extends HTMLAttributes<HTMLButtonElement> {}
+
+interface State {}
+
+class AdBlockDetector extends React.Component<Props, State> {
```
unjapones: Just to save a couple of lines, you can set this to be:
```typescript
class AdBlockDetector extends React.Component<Props, {}> {
```
... and remove line 5.

src/pages/marketplace.tsx
Agupane: Done, thanks!

src/components/common/adblock_detector.tsx
Agupane: I removed the reload and changed the message, But nevertheless I'll wait the response of @pablofullana to close this conversation

yarn.lock
Agupane: Was outdated, now It's ok, thanks!

src/components/common/adblock_detector.tsx
pablofullana: I support the no-reload approach.

About styling, I'd definitely go with a 'nicer' one. What about a pop up such as the one proposed in here: https://github.com/0xProject/0x-launch-kit-frontend/issues/118

Agupane: I'd added @gabitoesmiapodo in order to add those styles :)

PR #112, opened 2019-03-07
Basic notifications

### LINKED ISSUE #34: Track 'online' notifications
As a logged in user, I'd like to be notified every time an order but/sell order is executed/updated. 

**Update:** This version covers only those events that occur while user is 'online'. #106 deals with 'offline' events.

fvictorio:
Closes #34.

Show a notification when the user creates an order on limit, buys/sells on market, or cancels an order.

## Implementation

I added two things to the store: an array of notifications and a flag that indicates whether there are unread notifications. There's also an action that pushes a new notification: when this action is called, the new notification is added to the list and the unread flag is set to true at the same time.

Notifications are modeled with a discriminated union type. All notifications have a timestamp, that is used to show when the notification was sent, using [`react-timeago`](https://github.com/nmn/react-timeago. Then every notification sub-type may have specific fields. You'll notice that all of them have `amount` and `token` properties. I decided not to include those in the base type because we may add other notifications later that won't need them.


fvictorio:
@Agupane Done, thanks!

src/util/types.ts
```diff
@@ -98,3 +100,30 @@ export interface OrderBook {
     mySizeOrders: OrderBookItem[];
     spread: BigNumber;
 }
+
+interface BaseNotification {
+    timestamp: Date;
+}
+
+interface CancelOrderNotification extends BaseNotification {
+    kind: 'CancelOrderNotification';
```
unjapones: Non-blocking comment: we can use an enum (or string enum) for the kind attribute.

fvictorio: Done, thanks!

src/components/notifications/notifications_dropdown.tsx
```diff
@@ -0,0 +1,108 @@
...
+            <NotificationsDropdownHeader>
+                <BellIcon />
+                {hasUnreadNotifications ? <NewNotificationsBadge /> : null}
+            </NotificationsDropdownHeader>
+        );
+
+        const body = (
+            <NotificationsDropdownBody>
+                <NotificationDropdownTitle>Notifications</NotificationDropdownTitle>
+                <NotificationsList>{notificationsList}</NotificationsList>
```
unjapones: When there are no notifications, the dropdown's body opens anyway and shows a blank section.
I'm not 100% against it since (I believe) there are no designs for that scenario.

no_notifications

We can probably:
1. Keep it as it is,
2. Show a "No notifications" message inside the dropdown's body,
3. _Disable_ the dropdown by setting the `cursor: default;` + not showing the dropdown's body on click, 
4. Some other option that I can't think of right now.

I would pick option 2: I believe assigning a styled `<div>No notifications yet.</div>` to `notificationsList` if `notifications` is empty is enough (we'll have to ping @gabitoesmiapodo to check the style so it doesn't look _too bad_).

For option 3 I have an approach [here](https://github.com/unjapones/0x-launch-kit-frontend/blob/basic-notifications_disable-dropdown-when-no-notifications/src/components/notifications/notifications_dropdown.tsx), but I believe is too much.

unjapones: Also, if there are a few notifications (e.g. only 1), there is a white empty space on the dropdown's container.
If this is ok and expected, just ignore this comment.

empty_space

fvictorio: +1 to showing a message.

src/store/actions.ts
```diff
@@ -325,9 +362,27 @@ export const submitMarketOrder = (amount: BigNumber, side: OrderSide) => {
...
+            const txHash = await contractWrappers.exchange.batchFillOrdersAsync(
+                ordersToFill,
+                amounts,
+                ethAccount,
+                TX_DEFAULTS,
+            );
             dispatch(getAllOrders());
             dispatch(getUserOrders());
+
+            const tx = web3Wrapper.awaitTransactionMinedAsync(txHash);
```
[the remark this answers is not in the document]
fvictorio: Fixed, thanks!

src/components/notifications/notifications_dropdown.tsx
pablofullana: Couple of comments:
1. +1 to option 2: Show a "No notifications" message inside the dropdown's body.
1. Would it be too complex to limit dropdown size when there are just a few notifs?

gabitoesmiapodo: > Show a "No notifications" message inside the dropdown's body, 

+1

> Also, if there are a few notifications (e.g. only 1), there is a white empty space on the dropdown's container.

> Would it be too complex to limit dropdown size when there are just a few notifs?

Just change the list's container height from `height: 420px;` to `max-height: 420px;`

fvictorio: Done. @gabitoesmiapodo please take a quick look at 3254386 to see if that way of showing the "No notifications" message is good enough. Here's how it looks:

image

gabitoesmiapodo: It's fine, I think.

src/components/notifications/notification_item.tsx
```diff
@@ -0,0 +1,156 @@
...
+    public componentDidMount = () => {
+        const { item } = this.props;
+
+        if (item.kind === 'MarketNotification') {
+            this.setState({
+                active: true,
+                pending: true,
+            });
+
+            /* tslint:disable-next-line:no-floating-promises */
```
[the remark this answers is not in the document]
unjapones: I believe this should do the trick:
```typescript
try {
    await item.tx;
} finally {
    this.setState({ pending: false });
}
```
Source: https://dev.to/annarankin/finally-in-promises--trycatch-2c44 .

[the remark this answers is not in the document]
fvictorio: I left it as it was but added the `await`, that did the trick (I prefer it that way over the `try/catch`, since IMO the intention is clearer, but YMMV).

Issue #77, opened 2019-02-27
Put ETH price in store


fvictorio:
The USD price of ETH should be saved in the store and updated regularly. Components that use it should just get it from the store.

This means that this value will be nullable, since there's a time window —between the start of the app and the moment the price is obtained— in which the price is unknown. Also, the API could be down. So any component that depends on this should handle the case in which the value is not known.

_Originally posted by @Agupane in https://github.com/_render_node/MDI0OlB1bGxSZXF1ZXN0UmV2aWV3Q29tbWVudDI2MDc1OTc2Ng==/comments/review_comment_

PR #121, opened 2019-03-08
(Feature) Total costs calculation on LIMIT 

### LINKED ISSUE #101: Add "Total Cost" field of "Order Details" in buy/sell component (when in Limit)
Check #58 comments/updates.

Agupane:
Added UI implementation to display total cost on LIMIT

closes #101

### DIFF
--- a/src/components/marketplace/order_details.tsx
+++ b/src/components/marketplace/order_details.tsx
@@ -3,15 +3,35 @@ import React, { HTMLAttributes } from 'react';
 import styled from 'styled-components';
 
 import { MAKER_FEE } from '../../common/constants';
-import { getZeroXPriceInUSD } from '../../util/market_prices';
+import { getEthereumPriceInUSD, getZeroXPriceInUSD, getZeroXPriceInWeth } from '../../util/market_prices';
 import { themeColors, themeDimensions } from '../../util/theme';
+import { tokenAmountInUnitsToBigNumber } from '../../util/tokens';
+import { Token } from '../../util/types';
 import { CardTabSelector } from '../common/card_tab_selector';
 
 interface Props extends HTMLAttributes<HTMLButtonElement> {}
 
+enum OrderType {
+    Limit,
+    Market,
+}
+
 interface State {
-    orderDetailType: OrderDetailsType;
-    zeroXPriceInUSD: BigNumber;
+    limitOrder: {
+        orderDetailType: OrderDetailsType;
+        zeroXFeeInUSD: BigNumber;
+        zeroXFeeInWeth: BigNumber;
+        totalCostInWeth: BigNumber;
+        totalCostInUSD: BigNumber;
+    };
+    marketOrder: {};
+}
+
+interface Props {
+    orderType: OrderType;
+    tokenAmount: BigNumber;
+    tokenPrice: BigNumber;
+    selectedToken: Token | null;
 }
 
 const Row = styled.div`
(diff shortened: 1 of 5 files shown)


src/util/tokens.ts
```diff
@@ -5,6 +5,11 @@ export const tokenAmountInUnits = (amount: BigNumber, decimals: number, toFixedD
     return amount.div(decimalsPerToken).toFixed(toFixedDecimals);
 };
 
+export const tokenAmountInUnitsToBigNumber = (amount: BigNumber, decimals: number): BigNumber => {
```
fvictorio: Req: `tokenAmountInUnits` could use this function and add the `toFixed` to the result.

src/util/market_prices.ts
```diff
@@ -32,6 +38,35 @@ export const getZeroXPriceInUSD = async (): Promise<BigNumber> => {
...
+            sellOrders = sellOrders
+                .filter(order => order.side === OrderSide.Sell)
+                .sort((o1, o2) => o1.price.comparedTo(o2.price));
+            /* Check if there is any order that could fill the needs of 0x */
+            const orderWithEnoughZeroX = sellOrders.find(sellOrder => {
+                return sellOrder.size.cmp(MAKER_FEE) === 1;
+            });
+            zeroXPrice = orderWithEnoughZeroX ? orderWithEnoughZeroX.price : zeroXPrice;
+        }
+        if (!zeroXPrice) {
```
fvictorio: Q: Is there an scenario where this variable is falsy? I don't see it.

src/util/market_prices.ts
```diff
@@ -32,6 +38,35 @@ export const getZeroXPriceInUSD = async (): Promise<BigNumber> => {
...
+        const knownTokens = getKnownTokens(networkId);
+        const zeroXToken = knownTokens.getTokenBySymbol('ZRX');
+        let sellOrders = await getAllOrdersAsUIOrders(zeroXToken);
+        if (sellOrders && sellOrders.length > 0) {
+            sellOrders = sellOrders
+                .filter(order => order.side === OrderSide.Sell)
+                .sort((o1, o2) => o1.price.comparedTo(o2.price));
+            /* Check if there is any order that could fill the needs of 0x */
+            const orderWithEnoughZeroX = sellOrders.find(sellOrder => {
+                return sellOrder.size.cmp(MAKER_FEE) === 1;
```
fvictorio: Req: This is the same than using `greaterThan`, right? If that's the case:

- Using that function should be clearer.
- Shouldn't the comparison be greater than or equal to?

src/util/market_prices.ts
```diff
@@ -32,6 +38,35 @@ export const getZeroXPriceInUSD = async (): Promise<BigNumber> => {
...
+export const getZeroXPriceInWeth = async (): Promise<BigNumber> => {
+    /* Gets all the available sell orders of 0x and takes the lowest price, is there are no 0x sells orders available, the price is calculated using coin market */
+    const web3Wrapper = await getWeb3Wrapper();
+    let zeroXPrice = new BigNumber('0');
+    if (web3Wrapper) {
+        const networkId = await web3Wrapper.getNetworkIdAsync();
+        const knownTokens = getKnownTokens(networkId);
+        const zeroXToken = knownTokens.getTokenBySymbol('ZRX');
+        let sellOrders = await getAllOrdersAsUIOrders(zeroXToken);
+        if (sellOrders && sellOrders.length > 0) {
```
fvictorio: Q: Can `sellOrders` be falsy?

src/util/market_prices.ts
```diff
@@ -32,6 +38,35 @@ export const getZeroXPriceInUSD = async (): Promise<BigNumber> => {
...
+            sellOrders = sellOrders
+                .filter(order => order.side === OrderSide.Sell)
+                .sort((o1, o2) => o1.price.comparedTo(o2.price));
+            /* Check if there is any order that could fill the needs of 0x */
+            const orderWithEnoughZeroX = sellOrders.find(sellOrder => {
```
fvictorio: Q: What is the reason for this? In the (unlikely) case where the cheapest sell order has a size that is less than the fee amount, why wouldn't that price be a good enough estimation?

src/util/market_prices.ts
```diff
@@ -32,6 +38,35 @@ export const getZeroXPriceInUSD = async (): Promise<BigNumber> => {
...
+                return sellOrder.size.cmp(MAKER_FEE) === 1;
+            });
+            zeroXPrice = orderWithEnoughZeroX ? orderWithEnoughZeroX.price : zeroXPrice;
+        }
+        if (!zeroXPrice) {
+            const zeroXInUSD = await getZeroXPriceInUSD();
+            const ethInUSD = await getEthereumPriceInUSD();
+            zeroXPrice = zeroXInUSD.mul(ethInUSD);
```
fvictorio: Q: How does this work? It seems like it's multiplying a price in USD by a price in USD, does that make sense?

src/components/marketplace/order_details.tsx
```diff
@@ -70,18 +90,120 @@ class OrderDetails extends React.Component<Props, State> {
...
+                    limitOrder: {
+                        ...this.state.limitOrder,
+                        zeroXFeeInWeth,
+                        zeroXFeeInUSD,
+                        totalCostInWeth,
+                        totalCostInUSD,
+                    },
+                });
+            } catch (error) {
+                throw error;
```
fvictorio: Req: Catching and rethrowing an error is the same as not including the try/catch, as far as I know.

src/components/marketplace/order_details.tsx
```diff
@@ -70,18 +90,120 @@ class OrderDetails extends React.Component<Props, State> {
...
+        },
+    };
+
+    public updateLimitOrderState = async () => {
+        const { tokenAmount, tokenPrice, selectedToken } = this.props;
+        if (selectedToken) {
+            /* Reduces decimals of the token amount */
+            const tokenAmountConverted = tokenAmountInUnitsToBigNumber(tokenAmount, selectedToken.decimals);
+            /* This could be refactored with promise all  */
+            const promisesArray = [getZeroXPriceInWeth(), getZeroXPriceInUSD(), getEthereumPriceInUSD()];
```
fvictorio: Q: Am I missing something here or does this mean that we are going to get all the ZRX orders from the relayer every time the user presses a key?

Even if that's not a problem, shouldn't we handle the ZRX/wETH price the same way we handle the ETH/USD price: asking for it in regular intervals and having it as part of the global state?

src/components/marketplace/order_details.tsx
```diff
@@ -70,18 +90,120 @@ class OrderDetails extends React.Component<Props, State> {
...
-        const { orderDetailType } = this.state;
+        const { orderType } = this.props;
+        let render = null;
+        if (orderType === OrderType.Limit) {
+            const { orderDetailType, zeroXFeeInUSD, totalCostInWeth, totalCostInUSD } = this.state.limitOrder;
+            render = this._renderOrderDetails(orderDetailType, zeroXFeeInUSD, totalCostInWeth, totalCostInUSD);
+        }
+        if (orderType === OrderType.Market) {
+            const { orderDetailType, zeroXFeeInUSD, totalCostInWeth, totalCostInUSD } = this.state.marketOrder;
+            render = this._renderOrderDetails(orderDetailType, zeroXFeeInUSD, totalCostInWeth, totalCostInUSD);
```
fvictorio: Req: This line and line 183 are equal, they could be outside the `if`s.

src/components/marketplace/order_details.tsx
Agupane: It's now done as a set interval cache, we should add it on the store, maybe on the issue #77 ? @mariano-aguero

src/components/marketplace/order_details.tsx
Agupane: I'd just changed it

src/components/marketplace/order_details.tsx
Agupane: I'd removed it.

src/util/market_prices.ts
Agupane: Thanks for checking this, I'd made a mistake, it's a division what is needed to calculate that, now it's changed

src/util/market_prices.ts
Agupane: I'm not very sure if this is a good of wrong estimation. I'd discussed it with @unjapones and we decided to go for the "check the market price" approach, but we could just remove it and use the orderbook value. What do you guys think?

src/util/market_prices.ts
Agupane: I'd removed that check

src/util/market_prices.ts
Agupane: Done, now it's using that function and it's also checking the equality, thanks!

src/util/market_prices.ts
Agupane: I'd changed it now :)

src/util/tokens.ts
Agupane: Done!

PR #126, opened 2019-03-11
WiP: (Feature) Metamask locked - Metamask not installed

### LINKED ISSUE #116: Metamask locked
As a user, when having Metamask installed but locked, I'd like a get message stating so.

### LINKED ISSUE #117: Metamask not installed
As a user, when not having Metamask installed, I'd like to get a message stating so.

gabitoesmiapodo:
Closes #116 
Closes #117 

Styles for this should be done, but it needs some logic to properly work.

It shows error messages for both MM's _locked_ and _not installed_ status, on the sidebar and toolbar components.

I've implemented some (very) crude logic to display the different components, which can be used as follows:

**Toolbar:**

`src/components/common/toolbar.tsx`

Change _line 102_, `const mmLocked = false` to `const mmLocked = true` and you should see something like this.

Screen Shot 2019-03-11 at 15 25 41
    
Change _line 103_, `const mmNotInstalled = false` to `const mmNotInstalled = true` and you should see something like this.

Screen Shot 2019-03-11 at 15 25 15

**Sidebar:**

`src/components/marketplace/wallet_balance.tsx`

You can return three different values on `const getWalletStatus` (line 147).

Change its contents to `return WalletStatus.Ok;` and you'll get this for the "Wallet Balance" widget on the left

Screen Shot 2019-03-11 at 15 25 15

Change it to `return WalletStatus.NotConnected;` and you should get

Screen Shot 2019-03-11 at 15 48 13

And - finally- you can change it to `return WalletStatus.NoWallet;` to get

Screen Shot 2019-03-11 at 16 10 47


Agupane:
## Updated 13/03/2019
Added logic to detect if metamask is not installed or if the user did not granted the permissions

There is an open discussion about how the orderbook and orders UI should be displayed, right now it look like this:

image
image

If we decide keep this format, we will need to adjust the component to be centered

gabitoesmiapodo:
@Agupane 

It looks fine, I think.

If you want to center the error cards' text and icon, just add `textAlign="center"` when using it.

Not sure if centering it in the container is necessary.

mariano-aguero:
Hi @Agupane , there are some conflicts, could you solve them? Thank you

Agupane:
Done, thanks! @mariano-aguero

PR #131, opened 2019-03-12
(Feature) Total cost on market calculation

### LINKED ISSUE #103: Add "Total Cost" field of "Order Details" in buy/sell component (when in Market)

Agupane:
Calculates the total amount when the user buy/sells on market

closes #103

### DIFF
--- a/src/components/marketplace/order_details.tsx
+++ b/src/components/marketplace/order_details.tsx
@@ -3,15 +3,35 @@ import React, { HTMLAttributes } from 'react';
 import styled from 'styled-components';
 
 import { MAKER_FEE } from '../../common/constants';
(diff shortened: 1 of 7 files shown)


src/components/marketplace/order_details.tsx
```diff
@@ -280,4 +375,12 @@ class OrderDetails extends React.Component<Props, State> {
     };
 }
 
-export { OrderDetails };
+const mapStateToProps = (state: StoreState): StateProps => {
+    return {
+        state,
```
mariano-aguero: Is a good practice to map the entire state? What do you think @fvictorio @unjapones ?

PR #353, opened 2019-05-09
(Feature) Fixed size order book spread sticky centered

### LINKED ISSUE #335: Make the order book a fixed size, with the spread static and centered
Make the order book a fixed size, with the spread static and centered. (prevents spread from jumping around as orders come in)

See radar here for reference:

screen_shot_2019-04-25_at_4 18 04_pm

### LINKED ISSUE #359: Orderbook scroll performance
#353 adds some logic in the `onScroll` callback of the orderbook that makes it a little chumpy. I think the problem comes from using `setState` there.

gabitoesmiapodo:
Closes #335
Closes #359

Welcome to the _best of the bestest_ **sticky spread** implementation in existence. I hope you enjoy the ride.

**This should work as follows:**

1. Small amount of `Buy / Sell` items: it just gets centered along the Y axis:

Screen Shot 2019-05-09 at 14 29 35

2. Big amount of `Buy` or `Sell` items (enough to make the scrollbar appear): The items' list will scroll to bring the `Spread` row as close to the vertical center as possible.

If it's possible for the user to scroll enough as to make the `Spread` row go out of the container, it will stick to the top / bottom, depending on the scroll direction.

Screen Shot 2019-05-09 at 14 19 52

3. Large amount of `Buy` and / or `Sell` items (enough to make the scrollbar appear and for the container to scroll plenty):  The items' list will scroll until the `Spread` row is vertically centered.

The `Spread` row will also stick to the top / bottom when scrolling takes it out of its container.

Screen Shot 2019-05-09 at 14 27 51

----

**For responsive views:**

The "sticky" part still works, but I decided to disable the auto centering feature. Why? As it's implemented right now it centered the whole section (not the list only), and to change that would mean to add a bunch of extra code and to make logic even more convoluted to achieve something that's not really prioritary right now (that is: the responsive version).

So: no.

It works just fine and it's usable.


src/components/marketplace/order_book.tsx
```diff
@@ -28,6 +28,104 @@ interface OwnProps {
...
+const GridRowSpread = styled(GridRow)<GridRowSpreadProps>`
+    ${props => (props.stickySpreadPosition && props.stickySpreadPosition === 'top' ? 'top: 29px;' : '')}
+    ${props => (props.stickySpreadPosition && props.stickySpreadPosition === 'bottom' ? 'bottom: 0;' : '')}
+
+    background-color: ${props => props.theme.componentsTheme.cardBackgroundColor};
+    flex-grow: 0;
+    flex-shrink: 0;
+    display: grid;
+    position: ${props => (props.stickySpreadPosition !== '' ? 'absolute' : 'relative')};
+    width: ${props => (props.stickySpreadWidth ? props.stickySpreadWidth : 'auto')};
```
Agupane: R: I think we should try to start using on cases like this the `default props`, please [check this](https://www.styled-components.com/docs/advanced). You can see an example on [this PR](https://github.com/0xProject/0x-launch-kit-frontend/pull/351/files#r282266883)

The code for this should be something like:
```
const GridRowSpread = styled(GridRow)<GridRowSpreadProps>`
    ${props => (props.stickySpreadPosition && props.stickySpreadPosition === 'top' ? 'top: 29px;' : '')}
    ${props => (props.stickySpreadPosition && props.stickySpreadPosition === 'bottom' ? 'bottom: 0;' : '')}
     background-color: ${props => props.theme.componentsTheme.cardBackgroundColor};
    flex-grow: 0;
    flex-shrink: 0;
    display: grid;
    position: ${props => (props.stickySpreadPosition !== '' ? 'absolute' : 'relative')};
    width: ${props => (props.stickySpreadWidth};
    z-index: 12;
`;
```
```
GridRowSpread.defaultProps = {
    stickySpreadWidth: 'auto',
};
```

gabitoesmiapodo:
@mariano-aguero 

> An small glitch, if the orderbook is empty the spinner is not shown centered

Should be centered now.

@mariano-aguero @Agupane 

The janky scrolling should be fixed (or at least be less noticeable) now.

**Note:** You _might_ notice some jaggy scrolling / sticking if you happen to be scrolling when the component updates its content every few seconds. It doesn't seem to be related to the sticky spread's behavior (It behaves the same with or without it), but I think the update is triggering a re render of the component, which causes this problem.

Or at least that's what I think is going on.

Agupane:
@gabitoesmiapodo Please solve the conflicts, thanks!

src/components/marketplace/grid_row_spread.tsx
```diff
@@ -0,0 +1,86 @@
...
+                </CustomTD>
+                <CustomTDLast as="div" styles={customTDLastStyles}>
+                    {spreadValue}
+                </CustomTDLast>
+            </GridRowSpreadContainer>
+        );
+    };
+
+    public updateStickSpreadState = (stickySpreadState: StickySpreadState) => {
+        if (stickySpreadState !== this.state.stickySpreadState) {
```
fvictorio: This `if` is not necessary; React won't re-render if the state didn't change.

fvictorio:
Added a commit that uses the correct type for a ref.

PR #410, opened 2019-05-20
(Feature) Percentage Spread

### LINKED ISSUE #409: (Feature) Spread percentage in orderbook
Show spread percentage next to the spread amount
image

Agupane:
Closes #409 

Before:
image
After:
image


fvictorio:
This swaps the "My Size" and "Price" columns, since that's the order in the design. It also makes more sense with how the spread is shown.

src/components/erc20/marketplace/order_book.tsx
```diff
@@ -132,7 +144,7 @@ class OrderBookTable extends React.Component<Props> {
                                     borderTop: true,
                                 }}
                             >
-                                {spread.toFixed(UI_DECIMALS_DISPLAYED_PRICE_ETH)}
+                                {percentageSpread.toFixed(0)}%
```
fvictorio: Req: I think 0 decimal places won't make sense most of the time. I think 2 should work for the time being.

src/store/selectors.ts
```diff
@@ -134,17 +134,32 @@ export const getSpread = createSelector(
...
+    },
+);
+
 export const getOrderBook = createSelector(
     getOpenSellOrders,
     getOpenBuyOrders,
     getMySizeOrders,
     getSpread,
-    (sellOrders, buyOrders, mySizeOrders, spread): OrderBook => {
+    getSpreadInPercentage,
```
fvictorio: Req: These two selectors are not being used.

src/store/selectors.ts
```diff
@@ -134,17 +134,32 @@ export const getSpread = createSelector(
...
+    getOpenSellOrders,
+    (buyOrders, sellOrders) => {
+        if (!buyOrders.length || !sellOrders.length) {
+            return new BigNumber(0);
+        }
+
+        const lowestPriceSell = sellOrders[sellOrders.length - 1].price;
+        const highestPriceBuy = buyOrders[0].price;
+        const spread = lowestPriceSell.minus(highestPriceBuy);
+        return spread.dividedBy(lowestPriceSell).multipliedBy(100);
```
fvictorio: PP: This has a lot of duplication with `getSpread`. Maybe we can compute the percentage from the `getSpread` and `getOpenSellOrders` selectors? Or just make one selector that returns a pair?

src/components/erc20/marketplace/order_book.tsx
Agupane: Done

src/store/selectors.ts
Agupane: I'd removed them, thanks!

src/store/selectors.ts
Agupane: I'd used the `getSpread` and `getOpenSellOrders`, please check it again. I think making one selector that returns a pair it's not too easy to read as having just two selectors for different things

unjapones (approved):

mariano-aguero (approved):
LGTM, tested on ganache

Issue #321, opened 2019-04-24
Add filter control to the collections pages


fvictorio:
Depends on #319.

Add a control to filter the assets. You can show all the assets, only those in declining auction mode, or only those with a fixed price.

This control should be available both in the Collection page and in the My Collectibles page.

It's not _technically_ a requirement for this to be part of the URL, but since #320 will probably do it, we should do it here too for consistency.

Issue #320, opened 2019-04-24
Add sorting control to the collections pages


fvictorio:
Depends on #319.

Add a control to sort the assets by Newest added, Lowest price, Highest price and Most traded.

This control should be available both in the Collection page and in the My Collectibles page.

Note that #322 depends on this, and it implies that the sorting should be part of the URL (there are alternatives, but this seems the most logic option).

At this moment, it's not clear what "Most traded" and "Recently Listed" means, so that should be properly defined first.

## Notes

- For the "Most Traded" option, we can use the `num_sales` field returned by OpenSea, but I don't know if that will include sales done **outside** OpenSea
- For the "Recently Listed" option, we might use the salt field (that fiels is already used to store the creation date when you use the `ExchangeWrapper.cancelOrdersUpToAsync` method).

PR #418, opened 2019-05-21
Sort & filter collectibles (all/my collectibles) lists without URL params

unjapones:
Connects #320.
Connects #321.

### Includes

- Add `/src/util/sortable_collectibles.ts` to add the logic behind sorting `Collectibles` (Add `SortableCollectible`).
- Add `/src/util/filterable_collectibles.ts` in a similar fashion for the above.
- Modify `CollectiblesCardList`, `Props` so that it accepts a sort and filter criteria (to be used later in the landing page cc @mariano-aguero although you may need to add an optional argument to limit the number of items displayed).
- Add UI controls (dropdowns) specific to the filter and sort feature: Figma shows that they are different because of reasons.

### Does not include

- Set/get filter and sort criterias from the URL search params like `collectibles_sort=price_low_to_high&filter=dutch`.

### Screenshot

preview


src/components/erc721/collectibles/collectibles_list.tsx
```diff
@@ -27,32 +36,95 @@ const MainContainer = styled.div`
...
+        const { title } = this.props;
+        const collectibles = Object.keys(this.props.collectibles).map(key => this.props.collectibles[key]);
+        const { sortType, filterType } = this.state;
+
+        return (
+            <MainContainer>
+                <FiltersMenu>
+                    <Title>{title}</Title>
+                    <CollectiblesListSortStyled currentValue={sortType} onChange={this._onChangeSortType} />
+                    <CollectiblesListFilterStyled currentValue={filterType} onChange={this._onChangeFilterType} />
```
unjapones, line 120: Basically the `CollectiblesList` holds in its state the filter & sort values, and uses `CollectiblesListSortStyled` and `CollectiblesListSfiltertyled` to display and set them via a callback.

Then, passes the values of the sort, filter and collectibles to `CollectiblesCardList` to let it render accordingly.

src/util/sortable_collectibles.ts
```diff
@@ -0,0 +1,72 @@
...
+export interface SortableCollectible {
+    price: BigNumber | null;
+    creationDate: BigNumber | null;
+    collectible: Collectible;
+}
+
+export enum CollectibleSortType {
+    PriceLowToHigh = 'Price: low to high',
+    PriceHighToLow = 'Price: hight to low',
+    NewestAdded = 'Newest Added',
```
unjapones, line 15: I made the string values of the enum hold the text we are displaying in the UI. I'll probably change this later because it is generally not a very good idea. My suggestions is _do not add anything that depenes on the string value of this_.

mariano-aguero (approved):

src/components/common/button.tsx
```diff
@@ -23,6 +23,10 @@ const StyledButton = styled.button<{ variant?: string }>`
         props.variant && props.variant === 'tertiary'
             ? `background-color: ${props.theme.componentsTheme.buttonTertiaryBackgroundColor};`
             : ''}
```
Agupane: PP: Could you use default props here?, you can see an example [here](https://www.styled-components.com/docs/advanced). We are using the same in another part of the code (I did not found it sorry)

src/components/common/button.tsx
```diff
@@ -23,6 +23,10 @@ const StyledButton = styled.button<{ variant?: string }>`
...
+    ${props =>
+        props.variant && props.variant === 'quaternary'
+            ? `background-color: ${props.theme.componentsTheme.buttonQuaternaryBackgroundColor};`
+            : ''}
```
Agupane, line 29: PP: Hi, could you use default props?, [here](https://www.styled-components.com/docs/advanced) there is an example

Agupane (approved):

PR #423, opened 2019-05-22
Store collectibles list sort & filter values in URL search params

unjapones:
Connects #320.
Connects #321.

### Includes

- Move the texts of `CollectibleSortType` and `CollectibleFilterType` to their corresponding dropdowns UI components (previously, was using their string values, but now I use them as the values in the URL).
- Store collectibles list sort and filter values in the URL search params. Use `connected-react-router` action `replace` for this so the UI doesn't add history entries when changing the values.
- Fix `compareAscending` function in `util/sortable_collectibles.ts`: the "inverted" functions were putting the collectibles that lacked an order first.

### Does not include

- Spinner, yet.

### Preview

sort_filter

### DIFF
--- a/src/components/erc721/collectibles/collectibles_list.tsx
+++ b/src/components/erc721/collectibles/collectibles_list.tsx
@@ -1,8 +1,10 @@
+import queryString from 'query-string';
 import React from 'react';
 import { connect } from 'react-redux';
 import styled, { css } from 'styled-components';
 
-import { getOtherUsersCollectibles, getUserCollectibles } from '../../../store/selectors';
+import { setCollectiblesListFilterType, setCollectiblesListSortType } from '../../../store/actions';
+import { getOtherUsersCollectibles, getRouterLocationSearch, getUserCollectibles } from '../../../store/selectors';
 import { themeBreakPoints } from '../../../themes/commons';
 import { CollectibleFilterType } from '../../../util/filterable_collectibles';
 import { CollectibleSortType } from '../../../util/sortable_collectibles';
@@ -18,14 +20,15 @@ interface OwnProps {
 
 interface StateProps {
     collectibles: { [key: string]: Collectible };
+    search: string;
 }
 
-type Props = StateProps & OwnProps;
-
-interface State {
-    filterType: CollectibleFilterType;
-    sortType: CollectibleSortType;
+interface DispatchProps {
+    setSortType: (sortType: CollectibleSortType | null) => any;
+    setFilterType: (filterType: CollectibleFilterType | null) => any;
 }
+
+type Props = OwnProps & StateProps & DispatchProps;
 
 const MainContainer = styled.div`
     display: flex;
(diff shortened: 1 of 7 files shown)


src/components/erc721/collectibles/collectibles_list.tsx
```diff
@@ -101,16 +104,16 @@ const ButtonStyled = styled(Button)`
     }
 `;
 
-export class CollectiblesList extends React.Component<Props, State> {
-    public state = {
-        sortType: CollectibleSortType.NewestAdded,
-        filterType: CollectibleFilterType.ShowAll,
+export class CollectiblesList extends React.Component<Props, {}> {
+    public componentWillUnmount = () => {
```
unjapones, line 108: This removes the `sort` and `filter` search params from the URL when the components is unmounted.

src/util/sortable_collectibles.ts
```diff
@@ -10,20 +10,20 @@ export interface SortableCollectible {
...
+    NewestAdded = 'newest_added',
 }
 
 export const getCompareFunctionForSort = (sortType: CollectibleSortType) => {
     switch (sortType) {
         case CollectibleSortType.PriceLowToHigh:
-            return (a: SortableCollectible, b: SortableCollectible) => compareAscending(a.price, b.price);
+            return (a: SortableCollectible, b: SortableCollectible) => compareAscending(a.price, b.price, false);
         case CollectibleSortType.PriceHighToLow:
-            return (a: SortableCollectible, b: SortableCollectible) => compareAscending(a.price, b.price) * -1;
```
unjapones, line 23: The problem wih this is that the items that had `null` as the `price` (or `creationDate`) appeared at the beginning.

Agupane:
~~HI @unjapones, if one try to access the filtered-page directly from the url, the url gets removed and the user redirected to the default path~~

There was sort of non-updated build my side, it's working perfectly.

Agupane (approved):

src/components/erc721/collectibles/collectibles_list_sort.tsx
```diff
@@ -14,11 +14,11 @@ interface Props {
...
 
-const options: CollectibleSortType[] = [
-    CollectibleSortType.NewestAdded,
-    CollectibleSortType.PriceLowToHigh,
-    CollectibleSortType.PriceHighToLow,
-];
+const options: { [key: string]: string } = {
+    [CollectibleSortType.NewestAdded]: 'Newest Added',
+    [CollectibleSortType.PriceLowToHigh]: 'Price: low to high',
+    [CollectibleSortType.PriceHighToLow]: 'Pirce: high to low',
```
mariano-aguero: Req: typo

mariano-aguero (changes requested):

src/components/erc721/collectibles/collectibles_list_sort.tsx
unjapones: Good catch! :eye:

mariano-aguero (approved):

PR #426, opened 2019-05-22
(Feature) ERC20 Default theme as dark

### LINKED ISSUE #420: Set dark theme as the default for the erc20 frontend

Agupane:
Closes #420 
Adds support for having at the same time different themes for each app (ERC20 and ERC721). Each route has now a default theme (white for erc721 and dark for erc20).

Each application have their own `ThemeProvider` with their own theme


.env.example
```diff
@@ -12,3 +12,5 @@ REACT_APP_ENABLE_NO_METAMASK_PROMPT='true'
 REACT_APP_STEP_MODAL_DONE_STATUS_VISIBILITY_TIME='3500'
 REACT_APP_COLLECTIBLES_SOURCE='opensea'
 REACT_APP_COLLECTIBLE_NAME='CryptoKitties'
+REACT_APP_ERC20_THEME_NAME = 'WHITE_THEME'
+REACT_APP_ERC721_THEME_NAME = 'DARK_THEME'
```
unjapones: Req: I believe this should be `WHITE`, and the above line should be `DARK`

fvictorio:
The way this is implemented seems unnecessarily complex to me. Couldn't we just move the `ThemeProvider` inside each of the app components and use the proper constant there?

.env.example
Agupane: Yes, thanks for notice that, I'll make the changes

src/components/erc20/erc20_app.tsx
```diff
@@ -1,22 +1,56 @@
...
+                    <Route exact={true} path={`${ERC20_APP_BASE_PATH}/my-wallet`} component={MyWallet} />
+                </Switch>
+            </GeneralLayout>
+        </ThemeProvider>
+    );
+};
+
+const mapDispatchToProps = (dispatch: any) => {
+    return {
+        setThemeColor: (theme: Theme) => dispatch(setThemeColor(theme)),
```
Agupane: I think that now, having the theme on the store does not make sense any more, we could just remove it, or split it into two (one for erc20 and erc720). But It won't be used any more, so I think it's better to just remove it. Thoughs? @mariano-aguero @unjapones @fvictorio

Agupane:
I have made the refactor as you suggested @fvictorio, It was a good idea, it's better now, please check it

src/common/constants.ts
```diff
@@ -1,5 +1,10 @@
 import { BigNumber } from '0x.js';
 
+export enum MARKETPLACES {
+    ERC20 = 'ERC20',
+    ERC721 = 'ERC721',
+}
+
```
mariano-aguero: Req: I think this enum should not be here, Can we move it to the `util / types` ?

mariano-aguero (changes requested):
Req: I think we need to update the documentation related to `themes`

src/components/erc20/erc20_app.tsx
mariano-aguero: Mmmm I think it is not necessary, split it? neither, the theme is for the application(no matter what market you're ) and I think It should not be by market (there should not be two env vars for me)
Regards

src/common/constants.ts
Agupane: Done

Agupane:
The readme it's updated now @mariano-aguero

mariano-aguero (approved):

src/components/erc20/erc20_app.tsx
Agupane: As discussed internally, we decided to remove the themes from the store and two have two themes for the app.

unjapones (approved):

PR #431, opened 2019-05-22
Add collectibles list spinner

### LINKED ISSUE #320: Add sorting control to the collections pages

### LINKED ISSUE #321: Add filter control to the collections pages

unjapones:
Closes #320.
Closes #321.

### Includes

- Use spinner on the initial load of the landing page of the erc721 page and my collectibles.
- Show "No results" when sort/filter are applied in `collectibles_card_list.tsx` and there are no items to show.
- Appearance fixes on the landing page of the erc721 page (check late commits).
- Fix order of elements (buy collectible & description) on erc721 landing page when making the window narrower.

### Screen(s)

erc721's landing page:
preview


src/components/erc721/marketplace/sell_collectibles_button.tsx
```diff
@@ -6,6 +6,7 @@ import { Button } from '../../common/button';
 import { CollectibleListModal } from '../collectibles/collectible_list_modal';
 
 const ButtonStyled = styled(Button)`
+    min-width: 195px;
```
unjapones, line 9: Was a little bit narrower than what it is on Figma.

src/store/collectibles/reducers.ts
```diff
@@ -1,12 +1,13 @@
...
 
-import { Collectible, CollectiblesState } from '../../util/types';
+import { AllCollectiblesFetchStatus, Collectible, CollectiblesState } from '../../util/types';
 import * as actions from '../actions';
 import { RootAction } from '../reducers';
 
 const initialCollectibles: CollectiblesState = {
     collectibleSelected: null,
     allCollectibles: {},
+    allCollectiblesFetchStatus: AllCollectiblesFetchStatus.Request,
```
unjapones, line 10: Since we perform periodical refresh/updates, I thought about setting this as `Request` only on the initial state and show the loading/spinner in that moment.

src/store/collectibles/reducers.ts
```diff
@@ -16,7 +17,8 @@ export function collectibles(state: CollectiblesState = initialCollectibles, act
             action.payload.collectibles.forEach(collectible => {
                 allCollectibles[collectible.tokenId] = collectible;
             });
-            return { ...state, allCollectibles };
+            const allCollectiblesFetchStatus = AllCollectiblesFetchStatus.Success;
+            return { ...state, allCollectibles, allCollectiblesFetchStatus };
         case getType(actions.selectCollectible):
```
unjapones, line 22: `@TODO`: handle the error action type, which also implies to handle that on the UI. For a starter/base UI project I guess this is quite ok.

mariano-aguero (approved):

src/components/erc721/collectibles/collectibles_card_list.tsx
```diff
@@ -66,11 +72,20 @@ interface Props {
     sortType: CollectibleSortType;
     filterType: CollectibleFilterType;
     limit?: number;
+    isLoading?: boolean;
```
fvictorio, line 75: PP: I think this doesn't need to be optional (at least I removed the `?` and it compiled fine).

fvictorio (approved):

Agupane (approved):

src/components/erc721/collectibles/collectibles_card_list.tsx
unjapones: Yup. Made it optional in case you are feeding the component with a collection in a "sync way" and you don't want/need to signal any loading at all.

PR #438, opened 2019-05-23
(Feature) Adds a warning card if the user is not on an available network

### LINKED ISSUE #432: Show error state when connected to the wrong network
Screen Shot 2019-05-22 at 2 24 51 PM

This design can be found in the error states section on this figma file https://www.figma.com/file/hTFxa4hVhZtbqdLZwPgc8P/Launch-Kit-UI-erc20-new-colors-white?node-id=59%3A0

Agupane:
Closes #432 

Looks like this now:
image

### Description:
Basically uses the `Web3State.Error` to detect if the network id is not supported by the app and adds this check on both:
 `/src/components/erc20/marketplace/wallet_balance` and `/src/components/common/toolbar`

Note: the `Web3State.Error` could occur if the user is not on an available network, if there was an error trying to fetch account information **or** if there are not default contract addresses found for the given network id. Maybe we could split those cases to avoid showing a wrong error message for each case, but I think it's not very important right now (we are not doing anything with those states right now)

Based on those checks the components will render the following error card:
`<ErrorCard fontSize={FontSize.Large} text={errorsWallet.mmWrongNetwork} icon={ErrorIcons.Warning} />`

Note that this card should have the **IconComponent** for **Warning**, ~~but it is not available right now, later on @gabitoesmiapodo will add it~~. It's now available.


src/components/common/toolbar.tsx
```diff
@@ -100,7 +106,7 @@ const Toolbar = (props: Props) => {
                 <ErrorCard fontSize={FontSize.Large} text={errorsWallet.mmLoading} icon={ErrorIcons.Metamask} />
             ) : null}
             {!isMmLocked && !isMmNotInstalled && !isMmLoading ? <ToolbarCenter>{centerContent}</ToolbarCenter> : null}
-            {!isMmLocked && !isMmNotInstalled && !isMmLoading ? <ToolbarEnd>{endContent}</ToolbarEnd> : null}
+            {!isMmLocked && !isMmNotInstalled && !isMmLoading ? <ToolbarEnd>{endContentWrapper}</ToolbarEnd> : null}
```
mariano-aguero: Req: This code is duplicated with the line above, I think we can find a way to write it better

src/components/erc20/marketplace/wallet_balance.tsx
```diff
@@ -277,6 +285,32 @@ class WalletBalance extends React.Component<Props, State> {
...
+                    </WalletErrorFiller>
+                    <WalletErrorFiller top="0" right="0">
+                        {fillerBig()}
+                    </WalletErrorFiller>
+                    <WalletErrorFiller bottom="0" left="0">
+                        {fillerSmall()}
+                    </WalletErrorFiller>
+                    <WalletErrorFiller bottom="0" right="0">
+                        {fillerSmall()}
+                    </WalletErrorFiller>
```
mariano-aguero, line 309: Q: Why is this component repeated in this way?

mariano-aguero (changes requested):

src/components/erc20/marketplace/wallet_balance.tsx
gabitoesmiapodo: Those are four little gray graphics that serve not other purpose than to be some kind of placeholder / filler in fixed positions.

src/components/common/toolbar.tsx
Agupane: It's done

unjapones (approved):

PR #441, opened 2019-05-23
Add declining price chart to collectible details with recharts

mariano-aguero:
No issue related

#### Added:

- Recharts to the package json ([MIT License](https://github.com/recharts/recharts/blob/master/LICENSE) )
- Add a chart to a collectible if is dutch auction
Should look like this:

Screenshot_20190523_144610


mariano-aguero:
Hi @chriskalani 
This graph implementation uses [Recharts](http://recharts.org/en-US/), with a MIT LICENSE. Do you agree that we use this library? Any thoughts ? 

Regards

unjapones (approved):

src/components/erc721/common/declining_price_graph.tsx
```diff
@@ -0,0 +1,123 @@
...
+                        dataKey="markerLimit"
+                        stroke="#00AE99"
+                        fill="#00AE99"
+                        dot={{ stroke: '#00AE99', strokeWidth: 4 }}
+                    />
+                    <Area
+                        type="monotone"
+                        dataKey="currentPrice"
+                        stroke="#00AE99"
+                        fill="#00AE99"
```
Agupane: R: Could you put is values in as constants?

chriskalani:
Nice! Let's go with this

Agupane:
The chart looks awesome for me, I just added a comment, but it's non-blocking.

src/components/erc721/common/declining_price_graph.tsx
mariano-aguero: Added the color values to the theme properties

Agupane (approved):

Issue #393, opened 2019-05-17
Use TX_DEFAULTS with gasPrice


mariano-aguero:
Use always TX_DEFAULTS with gasPrice, make it mandatory (Maybe we can make a function that receives the gasPrice)

PR #448, opened 2019-05-24
Remove explicit gas limit

### LINKED ISSUE #446: Remove explicit gas limit
We are using a hardcoded gas limit in a lot of calls. This is a problem in mainnet since metamask shows an unreasonable number (above 1 USD for example) for transactions that shouldn't cost that.

This might cause problems in ganache when the estimated gas is off. If that happens, we can use this limit only during development.

mariano-aguero:
Closes #446 and #393

Added a function to get gas options


src/services/orders.ts
```diff
@@ -1,8 +1,8 @@
 import { assetDataUtils, BigNumber } from '0x.js';
 import { SignedOrder } from '@0x/connect';
 
-import { TX_DEFAULTS } from '../common/constants';
 import { getLogger } from '../util/logger';
+import { getGasOptions } from '../util/transactions';
```
fvictorio: PP: `getTransactionOptions`

fvictorio (approved):
Nice!

src/services/orders.ts
mariano-aguero: Done!

Agupane (approved):

src/store/blockchain/actions.ts
```diff
@@ -91,10 +92,7 @@ export const toggleTokenLock: ThunkCreator<Promise<any>> = (token: Token, isUnlo
...
                 ethAccount,
                 new BigNumber('0'),
-                {
-                    ...TX_DEFAULTS,
-                    gasPrice,
-                },
+                getTransactionOptions(gasPrice),
             );
         } else {
             tx = await contractWrappers.erc20Token.setUnlimitedProxyAllowanceAsync(token.address, ethAccount);
```
unjapones: Q: @mariano-aguero should we add the transaction options here too? Same question for line `161` in this file.

mariano-aguero: Yes, you can check this [right here](https://github.com/0xProject/0x-monorepo/blob/7423028fea4970306d7280bffa603844728d1431/packages/contract-wrappers/src/contract_wrappers/erc20_token_wrapper.ts#L236)
I will added in the next commit

unjapones (approved):

PR #457, opened 2019-05-27
Take collectible address from env

### LINKED ISSUE #456: Take collectible address from environment variable

fvictorio:
Closes #456.


src/common/constants.ts
```diff
@@ -81,11 +81,8 @@ export const COLLECTIBLES_SOURCE: string = process.env.REACT_APP_COLLECTIBLES_SO
 export const COLLECTIBLE_NAME: string = process.env.REACT_APP_COLLECTIBLE_NAME || 'Unknown';
 export const COLLECTIBLE_DESCRIPTION: string = process.env.REACT_APP_COLLECTIBLE_DESCRIPTION || 'Unknown';
 
-export const COLLECTIBLE_CONTRACT_ADDRESSES: { [key: number]: string } = {
-    1: '0xf5b0a3efb8e8e4c201e2a935f110eaaf3ffecb8d', // mainnet axie
-    4: '0x16baf0de678e52367adc69fd067e5edd1d33e3bf', // rinkeby cryptokitties
-    50: '0x07f96aa816c1f244cbc6ef114bb2b023ba54a2eb', // ganache mock erc721
-};
+export const COLLECTIBLE_ADDRESS =
+    process.env.REACT_APP_COLLECTIBLE_ADDRESS || '0x07f96aa816c1f244cbc6ef114bb2b023ba54a2eb';
```
Agupane: PP: Maybe adding two constants like: `RINKEBY_CRIPTOKITTIES` and `GANACHE_CRIPTOKITTIES` that we could use as `COLLECTIBLE_ADDRESS` would be better to let us test in dev

fvictorio: I though about that, but I ended up leaving it as it is, because the two main cases (at least for me) are ganache and mainnet. And if you want to test in the mainnet, you can use the address in `.env.example` (axis). To go back to ganache, you just comment out that variable.

That being said, if someone is regularly using Rinkeby CKs, we can add it as a variable or as a comment to make it more accessible.

unjapones (approved):

Agupane (approved):

PR #462, opened 2019-05-27
Add search collectibles modal

### LINKED ISSUE #325: Search assets
Clicking the search bar opens a modal where you can filter the assets with a search query. This list also has sorting and filter controls, so it might make sense to do this after #320 and #321 are done.

I don't know what should be used to match the query. I'm assuming that the name, or both the name and the description.

I also don't know what should be part of this list. I'm assuming that all collectibles: those that belong to the user (both for sale and not for sale) and those that other users are selling.

unjapones:
Closes #325.

### Includes

- Open a modal to search through the list of all assets in the redux state (search performed on name and ignored-case substring match). To open the modal, the user should click on the search input in the erc721 toolbar subapp.
- Make `collectible_card` a non redux-connected component and turn it a link/anchor (uses `Link` from `react-router-dom`
- Behavior was deduced from Figma's screenshots.
  - when is typed on the search bar, nothing is shown in the results.
  - when something is typed the results will appear accordingly and the text "Showing NN results" will render accordingly.
  - selecting a filter on the top-right corner affects the visibility of the results, but not the search result set (then, the text "Showing NN results" will not change).


src/components/erc721/modals/search-modal.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+    public render = () => {
+        const { isOpen, theme, allCollectibles } = this.props;
+        const { searchText, sortType, filterType } = this.state;
+
+        const collectibles = Object.keys(allCollectibles).map(key => allCollectibles[key]);
+        const searchResult = filterCollectibleByName(collectibles, searchText);
+        const shouldShowResults = searchText.length > 0;
+
+        Object.assign(theme.modalTheme.content, {
```
fvictorio: Req: This is modifying something that it's received via props. I think this may be the reason why the other modals look wider now.

fvictorio (changes requested):

fvictorio (approved):

unjapones:
The history of this PR/branch is a little bit weird, we might squash this when we decide it is finally ok.

Agupane (approved):

mariano-aguero (approved):

Issue #469, opened 2019-05-28
Reduce number of requests to relayer


fvictorio:
From @dekz:

> It appears that all markets are polled every interval, the other option would be to poll the current market displayed. When the interval is high this is a number of requests and was resulting in 429's (too many request errors) on Radar's Kovan endpoint

PR #473, opened 2019-05-28
(Feature) Split store update between apps

### LINKED ISSUE #402: updateStore thunk dispatches erc20 & erc721 thunks
The thunk `updateStore` from `/src/store/actions.ts` should have some logic to avoid fetching things that don't mind to the current sub-app (i.e. erc20 sub app doesn't carre about the `getAllCollectibles` thunk).

Agupane:
Closes #402 
Connects #469


src/store/actions.ts
```diff
@@ -18,6 +18,22 @@ export * from './market/actions';
...
+        currentRoute.includes(ERC20_APP_BASE_PATH) ? dispatch(updateERC20Store()) : dispatch(updateERC721Store());
+    };
+};
+
+export const updateERC721Store = () => {
+    return async (dispatch: any) => {
+        const web3Wrapper = await getWeb3Wrapper();
+        const [ethAccount] = await web3Wrapper.getAvailableAddressesAsync();
+        dispatch(getAllCollectibles(ethAccount));
+    };
```
unjapones, line 51: I believe we need the eth/weth balances too and update the gas info.

Agupane: It's done, please check it again

fvictorio: @Agupane I don't see any change related to @unjapones's comment.

fvictorio:
Maybe `initializeAppNoMetamaskOrLocked` should only be dispatched in ERC20?

src/store/actions.ts
Agupane: I removed         
```
dispatch(setEthBalance(ethBalance));
dispatch(setWethBalance(wethBalance));
dispatch(updateGasInfo());
dispatch(updateMarketPriceEther());
```
from `updateERC20Store` and it's now in `updateStore`, so they are dispatched for both apps (instead of duplicating that code on `updateERC20Store` and `updateERC721Store`)
@fvictorio

Agupane:
Yes sure, It's done now @fvictorio

src/store/blockchain/actions.ts
```diff
@@ -319,25 +315,14 @@ export const initWallet: ThunkCreator<Promise<any>> = () => {
...
-                    dispatch(setConnectedUserNotifications(ethAccount, networkId));
-                    // tslint:disable-next-line:no-floating-promises
-                    dispatch(updateMarketPriceEther());
-                } catch (error) {
-                    // Relayer error
-                    logger.error('The fetch orders from the relayer failed', error);
-                }
+                dispatch(updateMarketPriceEther());
+                // Inits wallet based on the current app
+                currentRoute.includes(ERC20_APP_BASE_PATH)
```
unjapones: Pp: it looks like this check is being used multiple times and is kind of important. Should we put this inside a function or selector? Lets' check what the other guys think about it with their :+1: or :-1:

unjapones: Also, I personally like to use `if` for statements (conditions/code flow) and the leave the `?` operator for expressions (assignment, return value), but that's being me picky.

mariano-aguero (approved):

unjapones (approved):

src/store/blockchain/actions.ts
fvictorio: Strong +1 to both things.

Agupane: It's done

PR #489, opened 2019-05-29
(Feature) Adds default name for collectibles

### LINKED ISSUE #477: Handle collectibles with null names
I just added a guard in 9990f6a22798086805892650269f59007897c062, but we should do a proper fix when we fetch the collectible info.

One option is to default to something like `${collectible.asset_contract.name} - #${collectible.token_id}`.

Agupane:
Closes #477 
Before:
image
After:
image

### DIFF
--- a/src/services/collectibles_metadata_sources/opensea.ts
+++ b/src/services/collectibles_metadata_sources/opensea.ts
@@ -16,7 +16,7 @@ export class Opensea implements CollectibleMetadataSource {
     public static getAssetAsCollectible(asset: any): Collectible {
         return {
             tokenId: asset.token_id,
-            name: asset.name,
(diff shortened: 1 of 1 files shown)


src/services/collectibles_metadata_sources/opensea.ts
```diff
@@ -16,7 +16,7 @@ export class Opensea implements CollectibleMetadataSource {
     public static getAssetAsCollectible(asset: any): Collectible {
         return {
             tokenId: asset.token_id,
-            name: asset.name,
+            name: asset.name ? asset.name : `${asset.asset_contract.name} - ${asset.token_id}`,
```
fvictorio:
```suggestion
            name: asset.name || `${asset.asset_contract.name} - #${asset.token_id}`,
```

mariano-aguero (approved):

PR #490, opened 2019-05-29
Don't prompt Connect MetaMask when env var is not set

### LINKED ISSUE #419: Metamask shouldn't prompt after every refresh
Every time the page reloads, it shows the user the metamask connection prompt. 
Screen Shot 2019-05-21 at 5 26 43 PM

unjapones:
Closes #419 .

What was happening: the default behavior, when the `.env` file lacked the the corresponding variable, was to always show the Connect MM modal (this was something desired at the beginning).

After some discussion, went ahead and removed the env var that controls the behavior of the "Connect MetaMask modal": now it will show only once, _and that's all_.


src/components/app.tsx
```diff
@@ -37,10 +37,10 @@ class App extends React.Component<Props> {
 
     public componentDidMount = () => {
         const wasWalletConnected = localStorage.getWalletConnected();
-        if (SHOULD_ENABLE_NO_METAMASK_PROMPT && wasWalletConnected) {
-            this.props.onConnectWallet();
-        } else {
+        if (SHOULD_ALWAYS_PROMPT_CONNECT_METAMASK || !wasWalletConnected) {
```
unjapones: Now that I think a little bit more about this:

Since this became super optional and not even listed in the `.env` file, _should we remove it and just leave the `!wasWalletConnected` check_?

Agupane: Yes I think it's better to just leave ` !wasWalletConnected`, makes more sense, at least for me

unjapones: Yeah, I agree too. If @fvictorio or @mariano-aguero add one :+1: more to this comment I'll go ahead and remove it.

fvictorio: :+1:

unjapones: Will go ahead as stated on the 1st comment in this thread and mark this as resolved.

mariano-aguero (approved):

fvictorio (approved):

PR #493, opened 2019-05-29
Add rate limiting for opensea and relayer

fvictorio:
Part of #469. I won't close that one yet because this can be further improved.

This is probably easier to review commit by commit. The one that makes `client` private is meant to make sure that every relayer request is rate limited.

### DIFF
--- a/src/services/collectibles_metadata_sources/index.ts
+++ b/src/services/collectibles_metadata_sources/index.ts
@@ -5,7 +5,7 @@ import { Mocked } from './mocked';
 import { Opensea } from './opensea';
 
 const sources: { [key: string]: CollectibleMetadataSource } = {
-    opensea: new Opensea(),
+    opensea: new Opensea({ rps: 5 }),
     mocked: new Mocked(),
 };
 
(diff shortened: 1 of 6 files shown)


Agupane (approved):

src/services/collectibles_metadata_sources/index.ts
```diff
@@ -5,7 +5,7 @@ import { Mocked } from './mocked';
 import { Opensea } from './opensea';
 
 const sources: { [key: string]: CollectibleMetadataSource } = {
-    opensea: new Opensea(),
+    opensea: new Opensea({ rps: 5 }),
```
mariano-aguero, line 8: PP/NB: maybe this rps property can be configurable

mariano-aguero (approved):

src/services/collectibles_metadata_sources/index.ts
unjapones: Mmmm initially thought the same, but I don't see anyone wanting to customize a lot this option. Also  this opensea metadata source is an example implementation so making it fully featured is not a priority right now.

unjapones (approved):

PR #495, opened 2019-05-29
(Feature) Now the fetch is in chunks of 10 orders

### LINKED ISSUE #488: Fetch collectibles info in bulk
This can be done using the [assets endpoint](https://docs.opensea.io/reference#getting-assets) and setting:

- `asset_contract_address` to the collectible address
- `token_ids` to the list of tokens to fetch

This should also be done in chunks of, say, 10. The docs don't specify a limit, but there's probably one (if not in the API, then in the server or the browser).

Agupane:
Closes #488


unjapones (approved):
:clap:

mariano-aguero (approved):

PR #500, opened 2019-05-30
(Feature) Close dropdown on click outside

### LINKED ISSUE #497: Clicking anywhere outside of the 721 balances dropdown should exit the dropdown
Screen Shot 2019-05-29 at 3 45 22 PM

right now it seems the only way to close it is by clicking the eth balance again

Agupane:
Closes #497 

## Description

Before this PR, the user needed to press the dropdown in the toolbar to close it, now the dropdown should close everytime the user clicks outside the dropdown (but only if the wETH converter is NOT open)

Also I'd made a little refactor of the `WalletConnectionStatusContainer` (see first commit). 
Now that component does not have status and is not responsible for set the `headerText`, it should be defined from the different `WalletConnectionContent` (ERC20 and ERC721)

### DIFF
--- a/src/components/account/wallet_connection_status.tsx
+++ b/src/components/account/wallet_connection_status.tsx
@@ -26,7 +26,7 @@ const WalletConnectionStatusText = styled.span`
 
 interface OwnProps extends HTMLAttributes<HTMLSpanElement> {
     walletConnectionContent: React.ReactNode;
-    shouldShowEthAccountInHeader: boolean;
+    shouldCloseDropdownOnClickOutside?: boolean;
     headerText: string;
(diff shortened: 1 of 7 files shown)


fvictorio:
This breaks the search in the market dropdown: when you click the input to type something, the dropdown is closed.

I don't think the `document.addEventListender('mousedown', ...` is the correct approach here. I think the usual way of doing this is with an overlay. There surely are other approaches, though.

Agupane:
It's fixed now @fvictorio

fvictorio (approved):

mariano-aguero (approved):

PR #502, opened 2019-05-30
Redirect to asset page after item is placed on sale

### LINKED ISSUE #496: Redirect to asset page after item is placed on sale
Screen Shot 2019-05-29 at 3 35 20 PM

mariano-aguero:
Closes #496

### DIFF
--- a/src/components/common/steps_modal/buy_sell_collectible_step.tsx
+++ b/src/components/common/steps_modal/buy_sell_collectible_step.tsx
@@ -2,6 +2,7 @@ import { BigNumber, SignedOrder } from '0x.js';
 import React from 'react';
 import { connect } from 'react-redux';
 
+import { STEP_MODAL_DONE_STATUS_VISIBILITY_TIME } from '../../../common/constants';
 import { getWeb3Wrapper } from '../../../services/web3_wrapper';
 import {
     createSignedCollectibleOrder,
@@ -15,6 +16,7 @@ import {
     getEthAccount,
     getStepsModalCurrentStep,
 } from '../../../store/selectors';
+import { sleep } from '../../../util/sleep';
 import {
     Collectible,
     OrderSide,
@@ -29,6 +31,7 @@ import { StepItem } from './steps_progress';
 
 interface OwnProps {
     buildStepsProgress: (currentStepItem: StepItem) => StepItem[];
+    closeModal: () => void;
 }
 interface StateProps {
     estimatedTxTimeMs: number;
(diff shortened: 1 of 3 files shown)


src/components/common/steps_modal/buy_sell_collectible_step.tsx
```diff
@@ -101,6 +113,11 @@ class BuySellCollectibleStep extends React.Component<Props, State> {
                 onLoading();
                 await this.props.submitCollectibleOrder(signedOrder);
                 onDone();
+
```
Agupane, line 113: R:
With the current logic, the user will be redirected to the collectibles page without the modal to be closed. I would add a "modalClose" function here before redirecting.

I think you can pass ` closeModal={reset}` to this modal as props  from `/src/components/common/steps_modals/steps_modals.tsx` and use that to close the modal before redirecting.

src/components/common/steps_modal/buy_sell_collectible_step.tsx
```diff
@@ -101,6 +113,11 @@ class BuySellCollectibleStep extends React.Component<Props, State> {
...
+                // Go to the collectible's profile
+                if (!currentRoutePath.includes(`collectible/${collectible.tokenId}`)) {
```
fvictorio: I think we need to stop doing this. We are hardcoding partial URLs everywhere and relaying a lot on this pattern of `currentRoutePath.includes`. It smells bad to me.

I don't know how is this supposed to be done the right way in `react-router`, but there must be a better way.

unjapones: Maybe we can dispatch [`goToIndividualCollectible`](https://github.com/0xProject/0x-launch-kit-frontend/blob/development/src/store/router/actions.ts#L70) and move this verification there?

To me, it seems that the condition check on the redirect is not a responsability of this component/file.

Agupane (changes requested):

src/components/common/steps_modal/buy_sell_collectible_step.tsx
mariano-aguero: Done in this [commit](https://github.com/0xProject/0x-launch-kit-frontend/pull/502/commits/64d5c2be5b301f3850e794aba827f4746b21edb9)

src/components/common/steps_modal/buy_sell_collectible_step.tsx
mariano-aguero: > Maybe we can dispatch [`goToIndividualCollectible`](https://github.com/0xProject/0x-launch-kit-frontend/blob/development/src/store/router/actions.ts#L70) and move this verification there?
> 
> To me, it seems that the condition check on the redirect is not a responsability of this component/file.

@unjapones  Done in this [commit](https://github.com/0xProject/0x-launch-kit-frontend/pull/502/commits/bc6b4686fcbe044c926f617111d27c157517a598)

fvictorio (approved):

Agupane (approved):

Issue #523, opened 2019-06-03
Setup default network


unjapones:
Currently there is a message "Wrong network switch to Mainnet" if the user is connected to Kovan (for example). 

The network should be configurable and constant: if the user changes to a network different than the one that was in the setup of the DApp, she/he should see a message in the UI to switch to the correct/configured on.

Issue #520, opened 2019-06-03
Orderbook doesn't show orders until wallet connects to Kovan


mariano-aguero:

unjapones:
This is kind of working ok but, if I'm on a network different than the one specified by the [`NETWORK_ID`](https://github.com/0xProject/0x-launch-kit-frontend/pull/533) the frontend doesn't seem to be fetching the corresponding orderbook.

PR #527, opened 2019-06-04
Orderbook doesn't show orders until wallet connects to Mainet

mariano-aguero:
Connects to #520 

I think there are two problems in the issue #520 :
- One is due to the use of a constant as a network id, it has to be solved in this issue #523 
- The other is because the application was trying to instantiate the web3Wrapper without being injected web3

This PR solves the second problem...

To test this problem it is necessary to set the MAINNET_ID constant in the network that needs to be tested and is configured in the relayer, check this [line](https://github.com/0xProject/0x-launch-kit-frontend/blob/development/src/store/blockchain/actions.ts#L477) related to this situation.

### DIFF
--- a/src/services/orders.ts
+++ b/src/services/orders.ts
@@ -31,6 +31,11 @@ export const getAllOrdersAsUIOrders = async (baseToken: Token, quoteToken: Token
     }
 };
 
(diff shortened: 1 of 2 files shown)


unjapones:
@mariano-aguero I changed the title of the PR because it was a little bit confusing: probably someone with MetaMask connected to Kovan would expect to see orders here, but that's not the case.

unjapones (approved):

Issue #503, opened 2019-05-30
ETH:wETH Converting error


burnsben:
ETH:wETH conversion is a bit jank at the moment

Getting this error after trying to wrap eth. 
<img width="413" alt="Screen Shot 2019-05-30 at 10 27 20 AM" src="https://user-images.githubusercontent.com/29830192/58651448-37094280-82c6-11e9-8ae4-37d78c373fd3.png">

Also this happened yesterday
<img width="490" alt="Screen Shot 2019-05-29 at 2 24 02 PM" src="https://user-images.githubusercontent.com/29830192/58651463-412b4100-82c6-11e9-98eb-375e4c79c061.png">

burnsben:
^for that second screenshot, those weren't my actual balances, it changed to 100 wETH randomly.

mariano-aguero:
Hi @burnsben, 
I think the problem is that previously there was an account change on metamask, and the account was not reset before the transaction, could you tell me if the problem persists? In case of Yes, could you share the steps to reproduce the issue?. 

Regards

![Screenshot_20190531_142801](https://user-images.githubusercontent.com/1144028/58723823-647eea80-83b1-11e9-92e5-fea031f8d19e.png)

burnsben:
Not sure what you mean by account change, the only changes I've made in metamask are switching from mainnet to kovan. 

Problems happen when I go back and forth between converting. 

Here, I only have 7.33 ETH, but it's showing an inflated total
<img width="354" alt="Screen Shot 2019-05-31 at 12 31 35 PM" src="https://user-images.githubusercontent.com/29830192/58730307-963b8580-83a0-11e9-84ff-cafced5c2b6b.png">

burnsben:
Going from full 7.33 ETH, to a split of ETH: wETH gave me this error again
<img width="347" alt="Screen Shot 2019-05-31 at 12 26 37 PM" src="https://user-images.githubusercontent.com/29830192/58730327-a7849200-83a0-11e9-99d1-0d863babe0a5.png">

burnsben:
wait sorry I got that reversed, that was trying to go from a split to full 7.33 ETH

burnsben:
the issue of showing the 10.3 amount is temporary-- there's a lag in calculating the individual amounts.

unjapones:
[...]

PR #528, opened 2019-06-04
Show proper message when user rejects/cancel ETH wrapping

unjapones:
Connects #503.

### Includes

- Show a proper message ('User denied transaction signature') when the user cancels/reject the wrap transaction in MetaMask. Previously, if the user rejected the transaction the message "`You have ${currentEthAmount} ETH but you need ${ethNeeded} ETH to make this operation`" would appear.
- Move and rename some exception files that existed on the `steps_modal` folder.

My suggestion is to check the 1st commit for the 1st item above, and the last one for the files move and rename.

### Screen

This is the screens of the corresponding message:

user_denied_transaction_signature


src/exceptions/user_denied_transaction_exception.ts
```diff
@@ -0,0 +1,7 @@
+export class UserDeniedTransactionSignatureException extends Error {
+    constructor(m: string = 'User denied transaction signature') {
```
mariano-aguero: R: Maybe we can use the default constant `USER_DENIED_TRANSACTION_SIGNATURE_ERR` right here

mariano-aguero (changes requested):

mariano-aguero (approved):

PR #534, opened 2019-06-07
Frontend supports only 18 decimal tokens

### LINKED ISSUE #516: Frontend supports only 18 decimal tokens
If your token has something else than 18 decimals you'll get weird calculations in orders and balances. My Wallet renders correctly. These orders fail, of course.

image

mariano-aguero:
Closes #516 

### Includes:

- Pass base token's decimals to `buy_sell.tsx`'s BigNumberInput, so the `amount` obtained is correctly expressed in base units.
- Fix `makerAssetAmount` calculation in the function `buildLimitOrder` from `src/util/orders.ts` (should be expressed in the taker asset's base units).
- Fix calculation of `price` in `buildLimitOrder` from `src/util/ui_orders.ts`.
- Adapt UI components that didn't considered decimals other than 18 to show a price (e.g. the markets dropddown).
- Fix failing tests due to changes in `src/util/orders.ts`.


unjapones:
Hey Mariano, I tested this yesterday and found out that there might be a problem withe the units handling. 
I created a `Sell` limit order with the following data:
- base token: `DGD, 9 decimals`
- quote token: `WETH, 18 decimals`
- value in "Amount" input: `2`
- value in "Price per token input: `1.5`

The above generated the following when MetaMask asked me to sign the order (screenshot):
wrong_signed_order_data

I believe there is a problem around the amounts and that the expected `SignedOrder` should end up with the following values:
```
makerAssetAmount: 2000000000,
takerAssetAmount: 3000000000000000000
```
`makerAssetAmount` should be expressed in base units of the maker asset, and the same goes for  `takerAssetAmount` (which has more zeros because WETH has more decimals).

Please confirm that you see the same thing. We'll continue improving this PR's branch.

mariano-aguero:
LGTM @unjapones , I think is ready to merge!

unjapones (approved):

PR #537, opened 2019-06-10
[WIP] Adding depth chart

dave4506:
TODO: 
- Set domain of graph dynamically based on order book
- Polish minor details and crashes

Screen Shot 2019-06-17 at 5 15 58 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
         "react-scripts": "3.0.0",
         "react-timeago": "^4.4.0",
         "react-tooltip": "^3.10.0",
+        "react-vis": "^1.11.7",
         "recharts": "^1.6.2",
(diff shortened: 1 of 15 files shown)


fragosti (commented):
This is just feedback from playing with it for a bit:

- I don't think the "volume" number is accurate, it seems to be off by 1000 for ZRX/WETH.
- The chart can sometimes cover the mid market price visual, but this might be fine.
- + / - should have hover: pointer.
- The popover is kind of stubborn and doesn't disappear as easily as I would expect.
- Tried switching to MKR/WETH and got a runtime error.
Screen Shot 2019-06-18 at 2 53 48 PM
Screen Shot 2019-06-18 at 2 55 36 PM

PR #546, opened 2019-06-19
Improvements in the conversion from eth to weth

mariano-aguero:
Connects #503

### Includes
- Covers these two problems described  by @burnsben 
    -- There's a lag in calculating the individual amounts.
    -- These errors happen after it says the transaction has completed successfully 
- This is a video reproducing the problem https://www.loom.com/share/7bd7fcd1c56c43198f2843eeaeb43aa1


src/store/blockchain/actions.ts
```diff
@@ -167,22 +171,36 @@ export const updateWethBalance: ThunkCreator<Promise<any>> = (newWethBalance: Bi
...
-                      balance: newWethBalance,
-                  }
-                : null;
-
-            dispatch(setWethTokenBalance(newWethTokenBalance));
-        });
+        await web3Wrapper.awaitTransactionSuccessAsync(tx);
+
+        const getBalanceInWei = async (): Promise<any> =>
+            retry(
```
unjapones: Some context about this addition:

>It looks like there is a small chance that the following occurs:
>- after a successful `contractWrappers.etherToken.depositAsync` (verified using `web3Wrapper.awaitTransactionSuccessAsync`)...
>- an immediate `web3Wrapper.getBalanceInWeiAsync(ethAccount)` may result in a value that does not reflectthe change in the balance.

unjapones (dismissed):

unjapones (approved):

PR #559, opened 2019-07-19
config - add missing dummy token address

adekbadek:
Without `dai` token address in config and using the local relayer (`REACT_APP_NETWORK_ID=50`), the app crashed:
```
Error: Token with symbol dai not found in known tokens
```
The token address was taken from [this docs page](https://github.com/0xProject/wiki/blob/master/developer-tools/3-Ganache-Setup-Guide.md#dummy-tokens-for-testing).


dekz:
Thanks @adekbadek

PR #560, opened 2019-07-19
layout - columns fix

adekbadek:
- Make columns scrollable. This fixes a bug which prevented "Buy" button from being clicked because it was covered by the footer:

before:
bug

after:
fix

- Wrap columns in a container to handle gutters in a universal way
- Refactor ColumnWide/ColumnNarrow to single Column component


dekz:
Thanks for the submission @adekbadek. It appears you cannot scroll now when the page is of very small resolution (but still desktop render).

adekbadek:
With this proposed solution the columns are scrollable:



But come to think of it, it might make more sense to just not force the footer to be at the bottom of the viewport and live with with the fact the it will not **always** be visible: 

sc

WDYT?

dekz:
@adekbadek hah I just had my cursor just a hair outside when I was testing it out. This looks fine to me, I'll take a deep look this week and hopefully get this in.

PR #564, opened 2019-07-29
Use MarketFill and remove orders which cannot be market filled

### LINKED ISSUE #554: Orderbook looks different before & after you connect your wallet
Wallet locked:

Screen Shot 2019-06-24 at 4 45 57 PM

**Wallet connected:**

Screen Shot 2019-06-24 at 4 46 06 PM

### LINKED ISSUE #563: Dust/grief orders are not removed
In the case where a number orders exist for $100 DAI and the maker has a balance of $30 DAI.

Only one of these orders can be `batchFilled` for $30 DAI. All remaining batch fills will fail.
None of these orders can be `marketFilled` as `marketFill` fills for the remaining amount ($100 DAI).

dekz:
Remove orders which cannot be filled for their entire remaining amount

fixes #563 
fixes #554

### DIFF
--- a/src/services/orders.ts
+++ b/src/services/orders.ts
@@ -23,8 +23,11 @@ export const getAllOrdersAsUIOrders = async (baseToken: Token, quoteToken: Token
     const orders: SignedOrder[] = await getAllOrders(baseToken, quoteToken);
     try {
         const contractWrappers = await getContractWrappers();
-        const ordersInfo = await contractWrappers.exchange.getOrdersInfoAsync(orders);
-        return ordersToUIOrders(orders, baseToken, ordersInfo);
+        const ordersAndTradersInfo = await contractWrappers.orderValidator.getOrdersAndTradersInfoAsync(
+            orders,
+            orders.map(o => o.makerAddress),
+        );
+        return ordersToUIOrders(orders, baseToken, ordersAndTradersInfo);
     } catch (err) {
         logger.error(`There was an error getting the orders' info from exchange.`, err);
         throw err;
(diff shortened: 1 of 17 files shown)


src/services/orders.ts
```diff
@@ -23,8 +23,11 @@ export const getAllOrdersAsUIOrders = async (baseToken: Token, quoteToken: Token
     const orders: SignedOrder[] = await getAllOrders(baseToken, quoteToken);
     try {
         const contractWrappers = await getContractWrappers();
-        const ordersInfo = await contractWrappers.exchange.getOrdersInfoAsync(orders);
-        return ordersToUIOrders(orders, baseToken, ordersInfo);
+        const ordersAndTradersInfo = await contractWrappers.orderValidator.getOrdersAndTradersInfoAsync(
```
fabioberger, line 26: This version of `OrderValidator` does not support MAP orders, and also does not validate signatures, FYI.

src/util/ui_orders.ts
```diff
@@ -36,19 +48,21 @@ const ordersToUIOrdersWithoutOrderInfo = (orders: SignedOrder[], baseToken: Toke
             filled,
             price,
             status,
+            makerFillableAmountInTakerAsset: order.takerAssetAmount,
+            remainingTakerAssetFillAmount: order.takerAssetAmount,
```
fabioberger, line 54: How come these are both assigned to `order.takerAssetAmount`?

dekz: This is the case where there is no wallet so no on-chain Provider exists in this case. We can only assume the total order is fillable and the maker can fill it entirely. Without a wallet the user isn't able to interact and this is reloaded in the alternative flow once a wallet is connected.

PR #574, opened 2019-11-01
Filter orders by custom maker addresses

BMillman19:
This feature adds the ability for the front-end to be filtered by a set of specified maker addresses. This filtering can be achieved by specifying a comma-separated list of ethereum addresses in the url via the `makerAddresses` query param.

To test:
- Copy `.env.example` into `.env`
- Update `REACT_APP_RELAYER_URL='https://api.radarrelay.com/0x/v2'`
- Update `REACT_APP_NETWORK_ID=1`
- Go to http://localhost:3001/#/erc20/?base=mkr&quote=weth&makerAddresses=0x2733bc229e016b743a086a66d92ecded9bf0b0fe

### DIFF
--- a/src/services/orders.ts
+++ b/src/services/orders.ts
@@ -12,15 +12,27 @@ import { getWeb3Wrapper } from './web3_wrapper';
 
 const logger = getLogger('Services::Orders');
 
-export const getAllOrders = (baseToken: Token, quoteToken: Token) => {
+const getAllOrders = async (baseToken: Token, quoteToken: Token, makerAddresses: string[] | null) => {
     const relayer = getRelayer();
     const baseTokenAssetData = assetDataUtils.encodeERC20AssetData(baseToken.address);
     const quoteTokenAssetData = assetDataUtils.encodeERC20AssetData(quoteToken.address);
-    return relayer.getAllOrdersAsync(baseTokenAssetData, quoteTokenAssetData);
+    const orders = await relayer.getAllOrdersAsync(baseTokenAssetData, quoteTokenAssetData);
+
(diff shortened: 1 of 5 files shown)


src/services/orders.ts
```diff
@@ -12,15 +12,27 @@ import { getWeb3Wrapper } from './web3_wrapper';
 
 const logger = getLogger('Services::Orders');
 
-export const getAllOrders = (baseToken: Token, quoteToken: Token) => {
+const getAllOrders = async (baseToken: Token, quoteToken: Token, makerAddresses: string[] | null) => {
```
steveklebanoff, line 15: Seems much more natural to have `makerAddresses?: string[]` - is there a reason reason to have the required param and `| null setup`?

steveklebanoff, line 15: It feels a little weird to have `getUserOrders` and then also `getAllOrders ` which takes in a list of `makerAddresses`. On first pass I would expect you to expand `getUserOrders` to take in multiple user addresses instead -- but I understand why you went about developing it this way

steveklebanoff (approved):

src/services/orders.ts
BMillman19: I did it this way because all of the fields in `MarketState` are defined definite as some time or `null`, but it looks like TS recognizes null as compatible with the `?` optional notation.

BMillman19: It's this way because `makerAddress` and other fields in the `MarketState` interface are defined as some type or `null`. This pattern of defining optional params is used all across the code base unfortunately.

